PR #12, opened 2016-04-14
Clarifai extractor

qmac:


tyarkoni:
Thanks! Maybe write a test for this when you have a chance.

PR #24, opened 2016-09-07
Add test for text dictionaries.

### LINKED ISSUE #20: Add tests of all remote dictionaries
We need some testing code that iterates over all dictionaries listed in [dictionaries.json](https://github.com/tyarkoni/featureX/blob/master/featurex/datasets/dictionaries.json) and makes sure they're all still available and work properly.

tsalo:
Should resolve #20.


tyarkoni:
Thanks!!

PR #36, opened 2016-09-08
Added SaliencyExtractor

shabtastic:
added SaliencyExtractor and associated tests; changed 'img' to 'stim' in extractors/image.py

### DIFF
--- a/featurex/extractors/image.py
+++ b/featurex/extractors/image.py
@@ -9,6 +9,7 @@ import time
 import numpy as np
 import tempfile
 import os
+import pySaliencyMap
 from warnings import warn
 
 # Optional dependencies
@@ -32,11 +33,11 @@ class BrightnessExtractor(ImageExtractor):
     def __init__(self):
         super(self.__class__, self).__init__()
 
-    def apply(self, img):
-        data = img.data
+    def apply(self, stim):
+        data = stim.data
         avg_brightness = np.amax(data, 2).mean() / 255.0
 
-        return Value(img, self, {'avg_brightness': avg_brightness})
+        return Value(stim, self, {'avg_brightness': avg_brightness})
 
 
 class SharpnessExtractor(ImageExtractor):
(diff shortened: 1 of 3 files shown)


featurex/extractors/image.py
```diff
@@ -9,6 +9,7 @@
 import numpy as np
 import tempfile
 import os
+import pySaliencyMap
```
tyarkoni: Can you add this to optional-dependencies.txt?

featurex/extractors/image.py
```diff
@@ -63,7 +64,35 @@ class VibranceExtractor(ImageExtractor):
...
+    def apply(self, stim):
+        # pySaliencyMap from https://github.com/akisato-/pySaliencyMap
+        data = stim.data
+
+        # Initialize variables
+        h, w, c = stim.data.shape
+        sm = pySaliencyMap.pySaliencyMap(h, w)
+
+        # Compute saliency maps and store full maps as derivatives
+        stim.derivatives = dict()
```
tyarkoni: Just making a note here: at some point, we'll probably want to add an economy flag to the package so that intermediate or derivative files aren't kept around if the user doesn't want them. We would then want to check against that flag here and conditionally store these derivatives.

featurex/tests/test_extractors.py
```diff
@@ -86,6 +87,15 @@ def test_part_of_speech_extractor():
     assert df.iloc[1, 3] == 'NN'
     assert df.shape == (4, 4)
 
+def test_saliency_extractor(self):
+        image_dir = join(_get_test_data_path(), 'image')
+        stim = ImageStim(join(image_dir, 'apple.jpg'))
+        tl = stim.extract([SaliencyExtractor()])
+        ms = tl.data['SaliencyExtractor'].data['max_saliency']
+        assert ((ms >= 0) & (ms <= 1))
```
tyarkoni: These assertions should probably be more specific, and check against the (presumed) correct value. If you're worried about precision issues, you can use numpy's `almost_equal` to make sure values are within N digits of each other.

featurex/extractors/image.py
shabtastic: done!

shabtastic: a note while we're figuring out the license issue: pySaliencyMap is not set up as a package in its current implementation.

tyarkoni:
I'll try to fix this up to pySaliencyMap with featureX this morning.

tyarkoni:
Merged, thanks!

PR #40, opened 2016-09-09
Add first pass at DerivedVideoStim.

tsalo:


featurex/stimuli/video.py
```diff
@@ -15,7 +16,14 @@ def __init__(self, video, frame_num, filename=None, data=None):
         self.frame_num = frame_num
         self.duration = 1. / video.fps
         self.onset = frame_num * self.duration
-
+    
+class VideoKeyframeStim(VideoFrameStim):
```
tyarkoni: Apparently "key frame" has a technical meaning in video encoding that's a little different from what we want. Any alternative names?

featurex/stimuli/video.py
```diff
@@ -1,8 +1,9 @@
...
+from __future__ import division
+from featurex.stimuli import DynamicStim
 from featurex.stimuli.image import ImageStim
 from featurex.core import Timeline, Event
-import six
 from moviepy.video.io.VideoFileClip import VideoFileClip
+import pandas as pd
 
 
 class VideoFrameStim(ImageStim):
```
tyarkoni: I think instead of adding a new `VideoKeyframeStim` class just to add a non-fps-based duration, we can allow the `VideoFrameStim` initializer to take an optional `duration` argument that defaults to `None`. Then if `duration=None`, we set to 1 / fps, just like the current approach, otherwise we use whatever was passed in as the value.

featurex/stimuli/video.py
tsalo: These frames are meant to represent following similar frames, so maybe "representative frame?"
Other ideas:
- Shot frame
- Scene frame

featurex/stimuli/video.py
tsalo: Just did this. I think it looks good.

tyarkoni:
Thanks!!

PR #41, opened 2016-09-09
fixed three bugs

emdupre:


tyarkoni:
Thanks! We may have to revisit the encoding issue, because that was supposed to be a fix for another encoding problem on Windows, but I'll merge for now.

PR #45, opened 2016-09-16
Unit tests

qmac:
Added unit test for the following:
-Brightness, Vibrance, Sharpness extractors
-Witai extractor
-Dense optical flow extractor

On this branch, with valid API keys, test coverage is 88%

Big items still uncovered:
-pySaliencyMap (22 lines)
-STFTExtractor with spectrogram flag on (15 lines)
-everything in featurex/datasets/text.py (35 lines)
-DerivedVideoStim (31 lines)


qmac:
Why coverage doesn't improve: travis still missing API keys, and I added more code that depends on the wit.ai API key, and coverage from the low-level image extractor tests only adds a few lines.

tyarkoni:
Looks good, thanks! I'll have to adapt these tests to the new API, but I'll merge as-is and worry about that later.

PR #46, opened 2016-09-16
Tesseract OCR

qmac:


featurex/tests/test_extractors.py
```diff
@@ -95,6 +96,14 @@ def test_saliency_extractor():
         sf = tl.data['SaliencyExtractor'].data['frac_high_saliency']
         assert np.isclose(sf,0.27461971)
 
+def test_tesseract_extractor():
+    pytest.importorskip('pytesseract')
```
tyarkoni: Is this easily installable from the command line? If so, can you add a line to the .travis.yml file that installs it in the install section? That way it'll get tested.

featurex/extractors/image.py
```diff
@@ -17,6 +17,15 @@
 except ImportError:
     pass
 
+try:
+    import pytesseract
```
tyarkoni: If the imported module is only called inside `apply`, we can probably move the import statement into the `apply` method. That way we also don't need to wrap it in a try/except block, since the tests won't run unless it's installed, and a user who tries to run it may as well see the exception so they know what's going on.

optional-dependencies.txt
```diff
@@ -4,3 +4,5 @@ matplotlib
 pysrt
 SpeechRecognition
 pySaliencyMap
+PIL
```
tyarkoni: I think PIL is superseded by pillow, which is already a hard dependency, so maybe see if you can use the Image class in pillow instead.

qmac: Ahh yes, it seems like pillow should work just fine.

featurex/extractors/image.py
qmac: Good point, I made the changes + for cv2 as well.

featurex/tests/test_extractors.py
qmac: Addressed: tesseract can easy be installed using `sudo apt-get install tesseract-ocr`

tyarkoni:
Haven't looked at this again yet, but maybe patch it up to work with the new API, and then I'll merge (and then we can repeat for the other branch).

tyarkoni:
Also looks like OpenCV has some dependencies that aren't being installed on travis. See [this line](https://travis-ci.org/tyarkoni/featureX/jobs/162304521#L564) in the log.

qmac:
Yea, might also be related to [this line](https://travis-ci.org/tyarkoni/featureX/jobs/162304521#L607) too.

tyarkoni:
The error in [this line](https://travis-ci.org/tyarkoni/featureX/jobs/162304521#L607) is due to a change in the opencv API from 2 --> 3, and I was already getting it locally (but the travis env was skipping the cv tests). I'm not entirely sure how we want to handle this, but it feels icky to start checking opencv versions in the code, so I think we'll just require opencv >= 3.

Issue #47, opened 2016-09-17
Values shouldn't be nested


tyarkoni:
When multiple Transformers are applied to a single Stim, the returned `Value` objects are nested, such that the keys in the top-level `Value.data` dict are `Transformer` names, and the values are other `Value` instances (whose `data` attribute is a normal dictionary of values). This is counter-intuitive and kind of horrendous. The returned top-level object should probably be either a plain `dict`, or some new container class (e.g., `ValueList`).

tyarkoni:
This is now obsolete as a result of #55.

PR #49, opened 2016-09-17
Core API refactoring and graph addition

tyarkoni:
WIP. A major restructuring of the core API plus addition of graph-based workflows.


tyarkoni:
This PR refactors much of the core API (and there are more major changes coming). Core changes include:

* The base `Extractor` class has been replaced in the hierarchy by `Transformer`, which is now the base class for any class that takes a `Stim` as input and transforms it into something else. Currently there are three `Extractor` subclasses:
    * `Filters` return a `Stim` of the same type as the input (e.g., applying some image filter to an `ImageStim`);
    * `Converters` return a `Stim` of a different type (e.g., pulling out an `AudioStim` from a `VideoStim`);
    * `Extractors` return some kind of data summary of features extracted fom a `Stim` (so effectively, what all of the current Extractors already do).
* The API has changed, so that all `Transformer` classes must now support a `transform` method. However, the way this is now handled internally is by calling a `_transform` method (which allows wrapping the class-specific call with additional logic). Each of the three `Transformer` subclasses must also implement a method corresponding to its class--i.e., every `Filter` implements `_filter`, every `Converter` implements `_convert`, and every `Extractor` implements `extract`.
* I've added a graph module that supports a functional (but currently pretty limited) mechanism for constructing arbitrary graphs of `Transformers`. This can be safely ignored for the moment; I plan to expand it soon.

There are a bunch of other minor tweaks and improvements. Some of the current changes may not make total sense, as they anticipate other forthcoming changes to the API. Most notably, as discussed in #47, there needs to be much more uniformity to the data types returned by `Extractor` classes. This is still a WIP though.

tyarkoni:
@qmac given the scope of the changes, it may make more sense for you to clone from the current master again and re-integrate the changes you've made than to try to resolve merge conflicts. Shouldn't be too much work; the main thing will be to update the new Extractors and tests to reflect the API changes.

PR #50, opened 2016-09-23
Google API

qmac:
Slight reworking of Google API extractors
Added logic for text, label, and property extractors


tyarkoni:
Looks great, thanks!

Issue #31, opened 2016-09-08
Identify key frames in videos based on magnitude of difference between frames


tyarkoni:
Many of the APIs only work on images, but we want to process videos by passing in individual frames. To keep processing efficient (and costs low for paid services), we want to pass in as few frames as we can get away with. Rather than processing every Nth frame, we could take the diff between every two frames and identify frames where the scene changes to a significant degree. This could be a method implemented in `VideoStim` that could be called by any API-based extractor that loops over frames.

tyarkoni:
Closing since the logic to do this was added in #53, though we still need to move this out of the `Stim` hierarchy and make it a `Converter`.

PR #53, opened 2016-10-09
Video stimuli keyframes

qmac:
Added unit tests for existing DerivedVideoStim class
Added functionality to filter frames using method outlined in #31 
Some refactoring


qmac:
TODO: implement as converter

PR #55, opened 2016-10-13
Revamped Extractor Results

tyarkoni:
Further changes to the core API, and some code reorganization:

* A new `ExtractorResults` class intended to replace `Value`, `Event`, and `Timeline`. This greatly simplifies working with `Extractor` results, at the cost of representing a bunch of NaN values for Extractors that don't return onsets and durations. See the `STFTExtractor` for a working example of the `ExtractorResults` in action. We will need to update all of the existing Extractors to work with this new approach.
* The base `Transformer` class and associated utilities are now in transformers.py (instead of core.py).


tyarkoni:
Tests will fail because the use of the new `ExtractorResults` class in `STFTExtractor` breaks a few tests, but I'm merging anyway, since we'll need to update all of the existing Extractors to use the same pattern anyway.

PR #58, opened 2016-10-14
Moving extractors to converters

qmac:
(1) Moved `DerivedVideoStim` `filter` functionality to a `FrameSamplingConverter`
(2) Moved `TesseractAPIExtractor` to a `TesseractAPIConverter` which inherits `ImageToTextConverter`
(3) Moved `GoogleVisionAPITextExtractor` to `GoogleVisionAPITextConverter` which also inherits `ImageToTextConverter` and required (4)
(4) Refactored core GoogleAPI interaction (working with credentials, executing requests, building requests for the Vision API) to a package-wide `GoogleAPITransformer` class, which can be subclassed by both `Extractors` and `Converters`.

### DIFF
--- a/featurex/converters/google.py
+++ b/featurex/converters/google.py
@@ -1,0 +1,41 @@
+import argparse
+import base64
+import os
+import re
+import sys
+from featurex.converters.image import ImageToTextConverter
+from featurex.converters import Converter
+from featurex.stimuli.image import ImageStim
+from featurex.stimuli.text import TextStim
+from featurex.google import GoogleVisionAPITransformer
+from featurex import Value, Event
+import tempfile
+from scipy.misc import imsave
+import numpy as np
+
+try:
+    from googleapiclient import discovery, errors
+    from oauth2client.client import GoogleCredentials
+except ImportError:
+    pass
+
+
+class GoogleVisionAPITextConverter(GoogleVisionAPITransformer, ImageToTextConverter):
+
+    request_type = 'TEXT_DETECTION'
+    response_object = 'textAnnotations'
+
+    def _convert(self, stim):
+        request =  self._build_request([stim])
+        responses = self._query_api(request)
+        response = responses[0]
+
+        if response:
+            annotations = response[self.response_object]
+            # Concatenate all the annotation
+            text = ''
+            for annotation in annotations:
+                text += annotation['description']
+            return TextStim(text=text)
+        else:
+            return TextStim(text='')
--- a/featurex/converters/google.py
+++ b/featurex/converters/google.py
@@ -25,6 +25,12 @@ class GoogleVisionAPITextConverter(GoogleVisionAPITransformer, ImageToTextConver
     request_type = 'TEXT_DETECTION'
     response_object = 'textAnnotations'
 
+
+    def __init__(self, handle_annotations='first', **kwargs):
+        super(GoogleVisionAPITextConverter, self).__init__(**kwargs)
+        self.handle_annotations = handle_annotations
+
+
     def _convert(self, stim):
         request =  self._build_request([stim])
         responses = self._query_api(request)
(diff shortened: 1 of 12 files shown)


featurex/converters/video.py
```diff
@@ -0,0 +1,79 @@
...
+            frame_index = video.frame_index
+        
+        if not hasattr(video, "history"):
+            history = pd.DataFrame(columns=["filter", "value", "n_frames"])
+        else:
+            history = video.history
+
+        if self.every is not None:
+            new_idx = range(video.n_frames)[::self.every]
+            history.loc[history.shape[0]]= ["every", self.every, len(new_idx)]
```
tyarkoni: Suppose you run a `DerivedVideoStim` through the converter again... let's say the old `history` has 1000 rows,  selected with `every`, but now we're only grabbing `top_n=100`. Then I think `history` will only have 100 rows updated, but the last 900 will persist untouched, even though they're no longer actually contained in the stim. Is that right? If so, I think we'll need to either figure out a more systematic way to track history, or just not persist `history` across stims.

For now I think the latter is probably the way to go. I do think at some point we should implement more systematic logging/history capabilities, but maybe the place to do that properly is in the context of the graph module. I.e., rather than creating a detailed history that's bound to each stim, we can create a `GraphLog` that compactly tracks provenance properly. One way to do that would be to reconstruct history on-the-fly from the graph rather than storing it. E.g., if we know that a `VideoStim` was fed to a `FrameSamplingConverter` twice in a row, with different options, then we can reconstruct the full history from the `.history` attribute of each `DerivedVideoStim`. But this is a low priority and for now I vote for just tracking only the proximal history inside each `Stim`.

featurex/converters/video.py
```diff
@@ -0,0 +1,79 @@
...
+        elif self.hertz is not None:
+            interval = int(video.fps / self.hertz)
+            new_idx = range(video.n_frames)[::interval]
+            history.loc[history.shape[0]] = ["hertz", self.hertz, len(new_idx)]
+        elif self.num_frames is not None:
+            import cv2
```
tyarkoni: This works fine, but at some point (not now) it might be worth looking to see if this could be efficiently implemented with pillow instead of opencv, since the former is much easier to install.

featurex/stimuli/video.py
```diff
@@ -72,57 +72,13 @@ class DerivedVideoStim(VideoStim):
     VideoStim containing keyframes (for API calls). Each keyframe is associated
     with a duration reflecting the length of its "scene."
     """
-    def __init__(self, filename, **kwargs):
+    def __init__(self, filename, elements, frame_index, history):
```
tyarkoni: We may want to allow `frame_index=None` and `history=None` here. Granted, a `DerivedVideoStim` created by a `FrameSamplingConverter` will always have a known `history`, but I can imagine a use case where someone might just want a way to indicate that something is derived from a `VideoStim` even if they aren't passing `frame_index` and `history` properly. Since we're currently not using those attributes for anything else, it seems safe to allow `None` values. If at some point we extend the API for this class to the point where `history` is required, it'll make sense to mandate those arguments.

featurex/converters/google.py
```diff
@@ -0,0 +1,41 @@
...
+        request =  self._build_request([stim])
+        responses = self._query_api(request)
+        response = responses[0]
+
+        if response:
+            annotations = response[self.response_object]
+            # Concatenate all the annotation
+            text = ''
+            for annotation in annotations:
+                text += annotation['description']
```
qmac: It doesn't do it in the API documentation examples, but it did in the 3 samples I tried. The common duplicate seems to be "<text>" and "<text>\n". I like the idea of the argument in `__init__` and I'll go ahead and code that up.

featurex/converters/video.py
qmac: @tyarkoni I'm not quite sure I understand the first paragraph. `history` just saves a row describing the operation, namely the name of the filter operation, the parameter of the filter, and how many frames are in `frame_index`. Either way, I agree with the idea that history should be something systematic for all transformers, in the graph module.

qmac:
Most comments addressed, with the exception of those for saving history in `DerivedVideoStim`

featurex/converters/video.py
tyarkoni: What I mean is that it's not clear why we check the `VideoStim` to see if it already has a `history` attribute (line 36). This would make sense if we want to essentially chain histories, but unless I'm mistaken, it's going to cause problems in cases where there already *is* a history, and we're now updating only some of the rows (per the `.loc` call, which will leave rows with non-overlapping existing index values untouched). My suggestion is to basically wipe the history each time (i.e., not to check for the video's existing history), so that all rows in the `.history` attribute only relate to the very last conversion. Does that make sense?

tyarkoni: Ohhhhh, right. Sorry, I wasn't reading that line properly. That makes sense. Merging!

PR #59, opened 2016-10-16
Diagnostics

qmac:
Implementation of the diagnostics module. The main purpose of the module is to provide the user with additional meta-information for newly extracted features, though the functionality can be applied to any Pandas DataFrame. The diagnostics are primarily focused on those relevant for multiple regression.

The diagnostics fall into three categories:
-collinearity: can one feature/column be linearly predicted from others, with good accuracy?
-outliers: which rows/columns are distant from their respective distributions?
-validity: are the features/columns valid/useful for regression?

The diagnostics are aggregated into `Diagnostic` objects, which exist in `featurex.diagnostics`.
This object runs each diagnostic test on a given DataFrame upon initialization. Moreover, `show()` (outputs diagnostic results) and `flag()` (gives back indices of concerning rows/columns) operations can be performed on the object.

Potential improvements: make flagging mechanism less hacky


featurex/diagnostics/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+        if isinstance(result, pd.DataFrame):
+            if diagnostic == 'CorrelationMatrix':
+                result = result.copy()
+                np.fill_diagonal(result.values, 0)
+            return result.applymap(thresh).sum().nonzero()[0]
+        else:
+            return result.apply(thresh).nonzero()[0]
+
+
+    def flag_all(self, thresh_dict={}):
```
tyarkoni: Nice, I like this idea. One minor tweak you could make that would make it a bit more powerful is to allow the user to indicate which diagnostics to use in the flagging. If you add optional `include` and `exclude` arguments that take lists, then the user can easily select only certain diagnostics to use.

featurex/diagnostics/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+        self.results['Eigenvalues'] = eigenvalues(self.data[cols])
+        self.results['ConditionIndices'] = condition_indices(self.data[cols])
+        self.results['VIFs'] = variance_inflation_factors(self.data[cols])
+        self.results['CorrelationMatrix'] = correlation_matrix(self.data[cols])
+        self.results['RowMahalanobisDistances'] = mahalanobis_distances(self.data[cols])
+        self.results['ColumnMahalanobisDistances'] = mahalanobis_distances(self.data[cols], axis=1)
+        self.results['Variances'] = variances(self.data[cols])
+
+
+    def show(self, stdout=True, plot=False):
```
tyarkoni: Maybe rename this `summary`, for consistency with other packages (e.g., statsmodels).

featurex/diagnostics/__init__.py
```diff
@@ -0,0 +1,123 @@
...
+                flagged = self.flag(diagnostic, thresh_dict[diagnostic])
+            else:
+                flagged = self.flag(diagnostic)
+
+            if diagnostic == 'RowMahalanobisDistances':
+                row_idx = row_idx.union(flagged)
+            else:
+                col_idx = col_idx.union(flagged)
+
+        return row_idx, col_idx
```
tyarkoni: We probably don't want to return sets, because I don't think they can be used in indexing arrays. Maybe convert to (sorted) lists on the return.

qmac:
I've addressed the suggestions in dfb9b5b

PR #60, opened 2016-10-19
IBM speech API

qmac:
Yay onsets!

### DIFF
--- a/featurex/converters/api.py
+++ b/featurex/converters/api.py
@@ -1,5 +1,7 @@
 import os
-from featurex.stimuli.text import ComplexTextStim
+import base64
+import json
+from featurex.stimuli.text import DynamicTextStim, ComplexTextStim
 from featurex.converters.audio import AudioToTextConverter
 
 try:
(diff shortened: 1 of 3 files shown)


tyarkoni:
Woah, awesome! Nice job!

featurex/converters/api.py
```diff
@@ -1,12 +1,21 @@
...
+import json
+from featurex.stimuli.text import DynamicTextStim, ComplexTextStim
 from featurex.converters.audio import AudioToTextConverter
 
 try:
     import speech_recognition as sr
 except ImportError:
     pass
 
+try:
```
tyarkoni: I think you can replace the urllib 2 vs. 3 imports with the version in `six.moves.urllib`, which wraps both versions in Python 3's API. We already have six as a dependency anyway, so I don't see a downside.

featurex/converters/api.py
```diff
@@ -1,12 +1,21 @@
 import os
-from featurex.stimuli.text import ComplexTextStim
+import base64
+import json
+from featurex.stimuli.text import DynamicTextStim, ComplexTextStim
 from featurex.converters.audio import AudioToTextConverter
 
 try:
     import speech_recognition as sr
```
tyarkoni: Even though it will increase redundancy slightly, we should probably import `sr` inside the initializers; that way we don't have to worry about catching exceptions.

PR #64, opened 2016-10-28
Updated extractors

qmac:
Updated api, google, text, and video extractors to use the new `ExtractorResult` format. Also made some fixes to get tests passing (included commenting out tests for old extractor format). Various other miscellany.

~~TODO (separate task/PR): remove all old extractor code~~ Addressed in #66

### DIFF
--- a/featurex/extractors/google.py
+++ b/featurex/extractors/google.py
@@ -1,6 +1,7 @@
 from featurex.extractors.image import ImageExtractor
 from featurex.stimuli.image import ImageStim
 from featurex.google import GoogleVisionAPITransformer
+from featurex.extractors import ExtractorResult
 from featurex import Value, Event
 import numpy as np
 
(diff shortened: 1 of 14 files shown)


featurex/extractors/google.py
```diff
@@ -1,49 +1,56 @@
...
+                                onsets=onsets, durations=durations)
 
 
 class GoogleVisionAPIFaceExtractor(GoogleVisionAPIExtractor):
 
     request_type = 'FACE_DETECTION'
     response_object = 'faceAnnotations'
-
-    def _parse_annotations(self, stim, annotations):
+    likelihood_dict = {'UNKNOWN': 0.0,
```
tyarkoni: Are these official values from Google? If not, I think we should probably stick with the strings... As much as I dislike the qualitative categories the API returns here, I think we want to avoid transforming values, as the user might not realize what's going on. A reasonable compromise would be to add a `convert_likelihood`  argument to the `transform` call that defaults to `False`, but does the conversion if the user explicitly sets `True`.

PR #66, opened 2016-11-05
Removed old core structure

qmac:
[Builds off commits in #64]

Went through and removed all the code and test dependencies on `Value/Event/Timeline` (in addition to the actual `core.py` implementations).

Some things are unsure of, such as cutting most of `Timeline`. Open to suggestions.


tyarkoni:
One minor issue here is that I think this PR reflects all of the commits in #66 except the last one--the numerical likelihoods are still there in the Face Extractor API. Do you want to take a quick look through this and make sure it's good to merge now that I've merged #66 into master?

PR #73, opened 2016-12-05
Implicit stimuli conversion

### LINKED ISSUE #48: Automatic Stim adapters
Consider a situation where a user wants to take a `VideoStim` as input and apply the `STFTExtractor` (i.e., short-time Fourier transform) to the audio track. Currently, an exception will be raised, because the `STFTExtractor` only handles `AudioStim` inputs. However, since most movies have an audio track, featurex should be smart enough to attempt to automatically extract an `AudioStim` from a `VideoStim` and apply the audioextractor to the result (i.e., basically building an implicit graph) before it raises an exception. This isn't a high priority, but would be nice to have at some point.

qmac:
An implementation of automatic stimuli conversion during validation of any transformation. Searches through available converters and uses one with the appropriate `_input_type` and `_output_type`, if one is found. Also added appropriate units tests.

Resolves #48

### DIFF
--- a/featurex/converters/api.py
+++ b/featurex/converters/api.py
@@ -14,7 +14,6 @@ from six.moves.urllib.error import URLError, HTTPError
 class SpeechRecognitionConverter(AudioToTextConverter):
     ''' Uses the SpeechRecognition API, which interacts with several APIs, 
     like Google and Wit, to run speech-to-text transcription on an audio file.
-    Defaults to using Wit.ai.
     Args:
         api_key (str): API key. Must be passed explicitly or stored in
             the environment variable specified in the environ_key field.
--- a/featurex/converters/api.py
+++ b/featurex/converters/api.py
@@ -1,7 +1,7 @@
 import os
 import base64
 import json
-from abc import abstractmethod
+from abc import abstractproperty
 from featurex.stimuli.text import TextStim, ComplexTextStim
 from featurex.converters.audio import AudioToTextConverter
 from featurex.converters.image import ImageToTextConverter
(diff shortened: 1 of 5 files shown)


featurex/tests/test_extractors.py
```diff
@@ -51,9 +50,39 @@ def test_implicit_stim_iteration():
...
                            columns='to', default_duration=1)
+    td = DictionaryExtractor(join(TEXT_DIR, 'test_lexical_dictionary.txt'),
+                             variables=['length', 'frequency'])
+    assert td.data.shape == (7, 2)
+    timeline = td.extract(stim)
+    result = timeline[2].to_df()
+    assert np.isnan(result.iloc[0, 1])
+    assert result.shape == (1, 4)
+    assert np.isclose(result['frequency'][0], 11.729, 1e-5)
+
```
qmac: Added back code for miscellaneous test. Got lost somewhere in 7fa2f767928eb90091148d25b390c151e3e8b55a.

featurex/transformers.py
```diff
@@ -54,6 +64,42 @@ def transform(self, stims, *args, **kwargs):
         return self._transform(self._validate(stims), *args, **kwargs)
 
 
+def get_converter(in_type, out_type):
```
tyarkoni: Did you try calling `get_transformer` with `base=Converter`? I haven't tested it, but in principle, it should work the same way.

featurex/converters/api.py
```diff
@@ -29,6 +30,7 @@ def __init__(self, api_key=None):
         self.recognizer = sr.Recognizer()
         self.api_key = api_key
 
+    @abstractmethod
```
qmac: This is to help https://github.com/tyarkoni/featureX/pull/73/files#diff-5bd92ddf7c5112f06b5c4d480ed14af3R93. Get converter will/should only give back non-abstract converters. SpeechRecognitionConverter shouldn't be initialized as a concrete class anyways, since it'd break without a valid `environ_key` or `recognize_method`.

qmac: Actually typing this out maybe making `environ_key` and `recognize_method` abstract would be the way to go

qmac:
Regarding using code reuse with `get_transformer`: the logic is pretty much the same except it checks the `_input_type` and `_output_type` as well as makes sure it is concrete. I suppose we could just use one method that enumerates the classes and have `get_transformer` and `get_converter` both call that method before checking their appropriate conditions

tyarkoni:
Yeah, just realized that and deleted that last comment. No worries!

featurex/converters/api.py
tyarkoni: That seems reasonable. Or maybe you could check inside `get_converter` if the class has a defined, non-`None` `environ_key` attribute.

qmac: I went with just making the properties abstract, seemed the most general way of doing this

PR #76, opened 2016-12-05
More efficient video frame handling

tyarkoni:
This PR eliminates automatic reading of all MoviePy VideoFileClip frames into memory when constructing a `VideoStim`. It adds a `get_frame` method for accessing specific frames (which can no longer be accessed through `.frames[i]`). Explicit conversion of movie frames to `VideoFrameStim` only happens when explicitly looping over a `VideoStim` now.


qmac:
Looks good to me, was there a particular use case this is addressing?

tyarkoni:
Yes, it turns out very large video files (@mic has some; see [studyforrest.org](studyforrest.org)) cause problems if we try to read in all frames on initialization. This way we can at least operate on individual frames without any problems (though many Converters and Extractors are still going to cause problems).

featurex/stimuli/video.py
```diff
@@ -12,7 +12,8 @@ class VideoFrameStim(ImageStim):
 
     def __init__(self, video, frame_num, duration=None, filename=None, data=None):
         self.video = video
-        self.data = self.video.frames[frame_num]
+        if data is None:
+            self.data = self.video.get_frame(index=frame_num)
```
qmac: @tyarkoni  This is problematic because `get_frame` returns a `VideoFrameStim` rather than image data. Should it return the data instead, or should this line have `.data` tacked to the end?

PR #78, opened 2016-12-06
Demo of a renderer for detected facial features.

mih:
At this point it cannot be considered "done". Due to #77 it can only handle cases with one detected image.

Also, a 3D head direction indicator located at the FOREHEAD_GLABELLA point would be great.

Confession: Google reports 3D coordinates of all facial feature locations. This code ignores Z completely. I should rather do a proper projection onto the image plane, but it is still unclear to me how Google wants to have the Z coordinate to be interpreted.

This is how it looks ATM:

test

### DIFF
--- a/featurex/support/render_face_features.py
+++ b/featurex/support/render_face_features.py
@@ -1,0 +1,128 @@
+from PIL import ImageDraw
+import numpy as np
+
+
+def _get_coord(d, label, nan=None):
+    if not nan:
+        nan = {}
+    # extract
+    coord = [d.get('{0}_{1}'.format(label, ax), None) for ax in ('x', 'y')]
+    # replace NaN if possible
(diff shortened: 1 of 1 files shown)


tyarkoni:
Looks neat, thanks! I'm going to leave this open for the moment; we'll be doing some restructuring and cleanup in the next few days, and then we'll figure out where to put this (and any future similar plotting functionality).

tyarkoni:
Now that we have something approaching a stable API for the core codebase, I'd like to merge this... but all the changes we've made in the meantime have broken some things. I'll try to patch this and merge this in the near future.

tyarkoni:
Unless you have objections, @mih, I'm going to close this for now. It would still be great to add something like this in the context of a `visualization` module, but that would require a bigger time investment than I think anyone has time for at the moment. Sorry!

PR #80, opened 2016-12-06
Test fixes

tyarkoni:
This PR fixes a couple of annoying issues that were preventing the tests from passing. All appears well now on both Python 2 and Python 3. I also got sidetracked and made a few other minor changes/tweaks, but nothing important.

PR #82, opened 2016-12-06
Propagate stimulus name during conversion

qmac:
When a stimulus is converted into another, tack on source information to the new stimulus.
Implemented with this in mind:
-`name` property holds what the stimulus should be called in an `ExtractorResult`
-`id` property holds a unique identifier for a stimulus, typically for those belonging to a `CollectionStimMixin`

There was some overlap with #80 but I don't think there will be any merge conflicts.


tyarkoni:
Looks good--merging!

Issue #88, opened 2016-12-13
Support other audio formats besides .wav when initializing AudioStim


tyarkoni:

tyarkoni:
Closed in #94.

PR #94, opened 2016-12-22
Using moviepy AudioClip to load in audio data

qmac:
Addresses #88 by using `moviepy` (already a requirement) to load in audio files.
Potential improvements:
-Support both mono and stereo data
-Modify the `VideoToAudioConverter` to simply use the VideoClip's `audio` property to construct a new `AudioStim` rather than writing a new audio file. This would require making sure no extractors or converters rely on the filename to work


tyarkoni:
Thanks @qmac--I think all of the improvements you suggest are well worth adding, but this addresses the basic issue, so I'm happy with it. Closing.

Issue #91, opened 2016-12-13
Add memoization of Converters


tyarkoni:
There will be a lot of overhead calling Converters repeatedly if implicit Stim conversion is required. We can address this by memoizing the conversion functions with joblib or something similar.

PR #95, opened 2016-12-22
Caching conversions

qmac:
Somewhat addresses #91 but it currently fails.

The problem with memoizing via joblib: we can’t pickle conversions of `VideoStim`s because pickling requires serializing `VideoClip`s which requires serializing `AudioClip`s. This doesn't work since `AudioClip`s have the property `make_frame` which is a lambda function. Unfortunately `pickle` does not like lambdas. This may be solved by using `dill`, but joblib seems to have stalled at implementing this (see https://github.com/joblib/joblib/pull/240).

There may be workarounds for this, or we can explore other caching libraries.

### DIFF
--- a/featurex/converters/__init__.py
+++ b/featurex/converters/__init__.py
@@ -1,12 +1,21 @@
 from abc import ABCMeta, abstractmethod, abstractproperty
 from featurex.transformers import Transformer, CollectionStimMixin
 from six import with_metaclass
+from tempfile import mkdtemp
+from joblib import Memory
+
+cachedir = mkdtemp()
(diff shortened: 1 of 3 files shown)


tyarkoni:
I ended up dropping and reconstructing the MoviePy `Clip` instances using the `__setstate__` and `__getstate__` magics. There should be very little overhead to re-initializing the `Clip`s, since the actual movie or audio data are read in separately (and can be serialized fine). So I think this is the path of least resistance. All existing tests pass, though I'm not sure if the memoization functionality is actually working as intended--will look into that now.

PR #112, opened 2017-01-10
Using io to open for encoding utf-8

adelavega:
Well I didn't mean to go down this path, but the downsampling of video was not going well on my measly laptop so I tried to do it on deepdream. 

For some reason, in windows when a file is in utf-8 encoding, that must be stated in `open`, and not in `json.load`. 

However, only python3 has different encodings in the `open` function. Turns out that you can make it python 2 and 3 compatible using open from the io library. Test seem to run well (aside from missing corpus ones for me).

### DIFF
--- a/pliers/datasets/text.py
+++ b/pliers/datasets/text.py
@@ -5,12 +5,12 @@ import os
 import tempfile
 import zipfile
 import shutil
-
+import io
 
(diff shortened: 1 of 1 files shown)


tyarkoni:
Thanks!

Issue #113, opened 2017-01-10
dictionary.json not found


adelavega:
When pliers is installed "traditionally" (aka python setup.py install), it is unable to find 'dictionary.json' during imports.

e.g.:
`IOError: [Errno 2] No such file or directory: '/Users/alejandro/anaconda2/lib/python2.7/site-packages/pliers-0.0.1-py2.7.egg/pliers/datasets/dictionaries.json'`

Notably, when using `python setup.py develop` this does not happen, so that may be why you missed it. I'm guessing that install only copies over .py files, whereas develop makes symlinks to the folders, and thus the file can be found.

I had no luck googling this issue.

tyarkoni:
Weird, I encountered this before and thought I fixed it. Will take a look. Thanks!

PR #124, opened 2017-01-19
Added datasets to package_data

adelavega:
In #113 package data was not being copied over during install in regular installs. 

Looks like the issue was `package_data` in `setup.py` was fetching pliers/data but dictionaries.json is in pliers/datasets.

This fixed it on osx, windows and linux for me.


tyarkoni:
Cool, thanks.

PR #136, opened 2017-02-24
Parallelization

### LINKED ISSUE #106: Add parallelization
Given the way we currently handle iterables in `Transformer`s, it should be pretty easy to add very basic support for embarrassingly parallel processing via `multiprocessing` or some similar module. E.g., given a list of `VideoFrameStim`s passed to a single `Extractor`, with no inter-frame dependencies, I see no reason why we couldn't launch multiple threads, one per `Stim`, to speed up processing.

tyarkoni:
Adds basic parallelization support (closes #106). Uses pathos instead of multiprocessing to avoid pickling-related issues (pathos uses dill). This adds a new dependencies, but since this is just extra (optional) functionality and pathos and its dependencies are pip-installable, this shouldn't be a problem.

I haven't tested to make sure it handles all kinds of `Stim` types, so that's probably worth doing at some point. But the existing tests  all pass with parallelization on, and many of them implicitly or explicitly use iterables, so it seems likely that everything will work okay.

Parallelization is controlled from the `config` module, where one can set `parallelize=True`, and also explicitly control the number of processes if desired (with `config.n_jobs`). By default, the number of processes will be set to cpu_count - 1, I think.


tyarkoni:
conda has been generating a crazy number of errors on Python 3 lately. These are usually resolved within a day or two via `conda update`, so I don't think there's any point in trying to debug. Tests all pass locally on 2/3 and remotely on 2, so I think we're good to merge despite the build failures.

pliers/transformers.py
```diff
@@ -13,6 +13,11 @@
 from copy import deepcopy
 import pandas as pd
 import os
+from functools import partial
```
qmac: Is this used anywhere?

tyarkoni: No, good catch. I also just realized my linter wasn't working properly, and fixed it. Will probably make a pass through the entire codebase to clean up other issues like this.

tyarkoni:
Good point re: parallelizing `run_node`; I'll take a look. One issue with trying to parallelize that is that because it's invoked recursively, there's a risk of running out of processes pretty quickly—especially if one of the branches involves an iterable (which will typically get assigned as many processes as are available). But I'm guessing that pathos is smart enough to not try to use more than one process if no more are available, so there may be a negligible cost to using the same naive implementation as for the iterables. I think we'll probably want more extensive tests for that, though.

PR #142, opened 2017-03-04
Support for remote stimuli

qmac:
After playing around with different approaches, I felt like it might be best to go with simply adding a `url` argument to all stim constructors, in addition to expanding the `load_stims` function. In this way URL's are treated exactly like the `filename` argument. This prevents the user from using some type of implicit conversion based on the http header in pipelines, but I figure the `load_stims` function should be ok. Thoughts?


tyarkoni:
If we go that route, do we even need a separate `url` field? Could we store urls in `filename` and just check for `http://` before loading?

tyarkoni:
Oh, I guess the main issue with this approach is the existing code will always try to initialize the `Stim` right away, including loading the data. So we can't have lazy loading of remote resources, I don't think...

tyarkoni:
On second thought (sorry for the stream-of-consciousness commenting), lazy loading is probably not something we need to be concerned with at the moment. On the assumption that users will typically want to do something with their `Stim`s immediately after loading, we may as well retrieve right away. This would only become an issue in the case of, e.g., social media posts that contain links we may want to defer processing of. But the way to handle that is probably to create some kind of deferred loader that doesn't even register the linked resources as `Stim` objects until there's some explicit call to that effect. So what you have here looks good.

qmac:
This is more/less independent of lazy loading right? If the particular `Stim` supports lazy loading via `filename`, it will still lazily load via `url`. Also, moviepy supports remote resources, so for video and audio stims its just simply setting `filename=url`.

tyarkoni:
Agreed, see above comment. I wasn't thinking of loading file or image data here, but more just recognizing that there are some potential `Stim`s that we don't want to initialize just yet. But that will only apply for things like tweets and facebook posts that contain links to images etc., which we don't support just yet anyway.

PR #145, opened 2017-03-05
Add Google SafeSearch Extractor

adelavega:
I added a new extractor to be able to query Google' SafeSearch, which detects if an image is NSFW (medical, adult, spoof, or violence). .


tyarkoni:
Thanks! Can you add a test?

tyarkoni:
@adelavega, @qmac pointed out something I missed: your test is failing because py.test should skip a test if it doesn't detect the `GOOGLE_APPLICATION_CREDENTIALS`. You have that line, but it should be `@pytest` and not `pytest`, so it's failing.

qmac:
I put the fix in #149

PR #147, opened 2017-03-06
Added resolution to IBM transcription

qmac:
IBM transcription will now result in `ComplexTextStim`s with `TextStim` elements that are either words or phrases (user provided argument to the converter). Previously, it partitioned word-by-word only.


tyarkoni:
Looks good, but can you modify the test for this `Extractor` to make sure it tests both settings of `resolution`?

qmac:
test added 👍

Issue #144, opened 2017-03-05
Add support for Indico Image services


adelavega:
Currently only text stimuli are supported but there are some nice image services

adelavega:
Closed by #148

PR #148, opened 2017-03-08
Indico image extractors

adelavega:
This is work in progress to address #144 

I've mostly focused on creating a base Indico class, and specialized Image and Text classes. Tests are passing for text. But I wanted to open this issue to discuss the fact that for Text, you can now choose the unit of analysis (e.g. word, or phrase). Ideally, we want to abstract away that piece of logic, so that the scorer/annotator can do its job for all. I suppose for VideoStims you could potentially also choose what the unit of analysis is, but right now that is being done by the implicit conversion to ImageStim at some sampling rate. 

Sounds like @qmac implemented it the same way for TextStim so we should be set.

### DIFF
--- a/pliers/extractors/__init__.py
+++ b/pliers/extractors/__init__.py
@@ -1,7 +1,7 @@
 ''' Extractor hierarchy. '''
 
 from .base import ExtractorResult, merge_results
-from .api import IndicoAPIExtractor, ClarifaiAPIExtractor
+from .api import IndicoAPITextExtractor, ClarifaiAPIExtractor
 from .audio import STFTAudioExtractor, MeanAmplitudeExtractor
 from .google import (GoogleVisionAPIFaceExtractor,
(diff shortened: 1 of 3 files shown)


adelavega:
OK this is done with the current structure. @tyarkoni @qmac please review (looks like I don't have the permissions to formally request one).

pliers/extractors/__init__.py
```diff
@@ -1,7 +1,7 @@
 ''' Extractor hierarchy. '''
 
 from .base import ExtractorResult, merge_results
-from .api import IndicoAPIExtractor, ClarifaiAPIExtractor
+from .api import IndicoAPITextExtractor, IndicoAPIImageExtractor, ClarifaiAPIExtractor
```
qmac: also update the `_all_` variable below

qmac:
Looks good to me

PR #149, opened 2017-03-09
Uses audio property of video to convert to audio

### LINKED ISSUE #141: Make VideoToAudioConverter simpler
Modify the `VideoToAudioConverter` to simply use the `VideoClip`'s audio property to construct a new `AudioStim` rather than writing a new audio file. This would require making sure no extractors or converters purely rely on the filename to work (at the very least if a filename is needed it will lazily write it out).

qmac:
Resolves #141

Requires every audio extractor that needs a file on disk to check if the `AudioStim` has a `filename`, and write a temporary one to disk if it doesn't.


tyarkoni:
Following on the above, I think a reasonable way to handle this is to have the `AudioStim` have a method that returns a context manager that automatically removes the file when the `with` block ends. That way both the creation and deletion of the tempfile are handled in one place.

If you don't have time to get to this before you leave, I'll just merge as-is and we can revisit this later.

PR #154, opened 2017-03-23
Moved temp file management to stim classes

qmac:
Many transformers require a local file to work. Previously, we just created a temp file on the fly in the transformer logic. This PR moves the logic to a `get_filename` method in the stim classes, to prevent code reuse.

### DIFF
--- a/pliers/converters/api.py
+++ b/pliers/converters/api.py
@@ -3,7 +3,6 @@
 import os
 import base64
 import json
-import tempfile
 from abc import abstractproperty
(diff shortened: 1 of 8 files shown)


qmac:
Yea I considered that but they also use different saving methods (`imsave` for images, `clip.write_audio_file` for audio, etc.) so it gets more complicated.

tyarkoni:
Ah, makes sense. Will merge it as-is then.

PR #165, opened 2017-04-09
Various fixes/slight enhancements

qmac:
- Keeping durations in graph results
- Including original `source_file` in graph results
- Allowing users to name nodes and transformers within a graph


tyarkoni:
Looks okay to merge, though we should probably get to the bottom of that test failure at some point.

PR #175, opened 2017-05-05
Sklearn integration

### LINKED ISSUE #139: scikit-learn integration
Per discussion the other day, we should add utilities to wrap `Transformer` and `Graph` classes in a lightweight class that supports the sklearn `TransformerMixin` (basically we just need to implement a `fit_transform` method that executes the pliers `Transformer` or `Graph`). This will make it super easy for people to embed complex feature extraction pipelines in their existing sklearn workflows.

qmac:
Added code for wrapping `Extractor` or `Graph` objects as scikit-learn `Transformer`s.

Includes:
-Implementation
-Tests
-Example notebook illustrating basic usage within a sklearn `Pipeline` in `examples/`

Resolves #139

### DIFF
--- a/optional-dependencies.txt
+++ b/optional-dependencies.txt
@@ -10,4 +10,4 @@ SpeechRecognition>=3.6.0
 pytesseract
 IndicoIo
 tensorflow>=1.0.0
-sklearn
+scikit-learn
--- a/optional-dependencies.txt
+++ b/optional-dependencies.txt
@@ -10,3 +10,4 @@ SpeechRecognition>=3.6.0
 pytesseract
 IndicoIo
 tensorflow>=1.0.0
+sklearn
--- a/pliers/scikit.py
+++ b/pliers/scikit.py
@@ -17,6 +17,7 @@ class PliersTransformer(BaseEstimator, TransformerMixin):
 
     def __init__(self, transformer):
         self.transformer = transformer
+        self.metadata = None
 
     def fit(self, X, y=None):
         return self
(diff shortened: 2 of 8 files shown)


optional-dependencies.txt
```diff
@@ -9,4 +9,5 @@ seaborn
 SpeechRecognition>=3.6.0
 pytesseract
 IndicoIo
-tensorflow>=1.0.0
\ No newline at end of file
+tensorflow>=1.0.0
+sklearn
```
tyarkoni: I think this should be `scikit-learn` here

pliers/scikit.py
```diff
@@ -0,0 +1,41 @@
+from pliers.extractors import Extractor, merge_results
+
+try:
+    from sklearn.base import TransformerMixin, BaseEstimator
+except:
+    TransformerMixin, BaseEstimator = None
```
tyarkoni: has to be `None, None` on right side, I think

pliers/scikit.py
```diff
@@ -0,0 +1,41 @@
...
+class PliersTransformer(BaseEstimator, TransformerMixin):
+
+    ''' Simple wrapper for using pliers within a sklearn workflow.
+    Args:
+        transformer (Graph or Transformer): Pliers object to execute. Can
+            either be a Graph with several transformers chained or a single
+            transformer.
+    '''
+
+    def __init__(self, transformer):
```
tyarkoni: We could also probably allow it to take a string as the `transformer` argument, in which case it initializes the corresponding `Transformer` with all default arguments. Could save user a line of code in many cases.

pliers/scikit.py
```diff
@@ -0,0 +1,41 @@
...
+        if isinstance(self.transformer, Extractor):
+            result = self.transformer.transform(stimulus_files).to_df()
+        else:
+            result = self.transformer.transform(stimulus_files, merge=False)
+            result = merge_results(result, flatten_columns=True)
+
+        extra_columns = list(set(['onset', 'duration', 'history', 'class',
+                                  'filename', 'stim_name', 'source_file']) &
+                             set(result.columns))
+        self.metadata = result[extra_columns]
```
tyarkoni: The sklearn convention for storing attributes estimated from data is to have names with trailing underscores (see [this](http://scikit-learn.org/0.16/developers/index.html#estimated-attributes)), so maybe call it `metadata_`.

PR #180, opened 2017-06-12
added google web entities extractor

andrewheusser:
This PR adds support for the google vision web detection API, which works similarly to the labels API, but the labels are derived from a web search.  Specifically, the extractor returns web 'entities' that were discovered as features (e.g. Barack Obama, White House) and values (.5, .3).

To implement this, I:

+ added a `GoogleVisionAPIWebEntitiesExtractor` to `pliers/extractors/google.py`
+ imported it into `pliers/extractors/__init__.py`
+ added a test in `tests/test_google_extractors.py::test_google_vision_api_web_entities`
+ made reference to the extractor in the API Keys section of the README

One outstanding question I have about the other extractors is how they handle errors.  In my implementation, I used a try/except statement to handle cases where no web entities were returned.  I'm not sure that this is the best way to handle this, so I wanted to bring it up.  See snippet below:

```
    def _parse_annotations(self, annotations):
        features = []
        values = []
        try:
            for annotation in annotations['webEntities']:
                features.append(annotation['description'])
                values.append(annotation['score'])
        except:
            pass
        return features, values
```
Happy to make change/additions, let me know!


qmac:
What error do you get without the `try/except`? Does it iterate over an empty list? If so I don't think there's anything wrong with returning empty feature/value lists.

andrewheusser:
I've encountered 2 different errors:
 + annotations['webEntities'] doesn't exist
 + annotation['description'] doesn't exist

and it crashes on these, which is why i added the try/catch.  

I could handle those cases more specifically if you'd like:

```
if 'webEntities' in annotations:
    for annotation in annotations['webEntities']:
        if 'description' in annotation and 'score' in annotation:
            features.append(annotation['description'])
            values.append(annotation['score'])
```

qmac:
Ah I see. I'm more in favor of the specific checking. I find it a little odd that annotation['description'] may not exist, as I never encountered such an error in the labels API, but that may require more testing/investigation on my part. Looks good to merge otherwise.

andrewheusser:
I found it strange as well.  Could have been a glitch in the matrix...or a glitch in my brain :). On the other hand, there is minimal computational cost to leave in the check. happy to take it out though.

qmac:
Looks good. Thanks!

andrewheusser:
👍

PR #181, opened 2017-07-08
Minor cleanup and improvements

qmac:
- Went through and made sure all `_log_attributes` were correct
- Went through and made all appropriate `Transformers` inherit the `EnvironmentKeyMixin`
- Upgraded the `ClarifaiAPIExtractor` to use v2 of the API
- Adjusted `TensorFlowInceptionV3Extractor` to output in the same format as other object recognition extractors, i.e. using objects as columns instead of the `label_1` `score_1` format
- Fixed bug in stemming filter by removing the `WordNetLemmatizer` since it does not have a `.stem` method

### DIFF
--- a/pliers/extractors/api.py
+++ b/pliers/extractors/api.py
@@ -11,6 +11,7 @@ from pliers.extractors.image import ImageExtractor
 from pliers.extractors.text import TextExtractor
 from pliers.extractors.base import Extractor, ExtractorResult
 from pliers.transformers import BatchTransformerMixin, EnvironmentKeyMixin
+from pliers.utils import listify
 
 
 try:
@@ -158,6 +159,7 @@ class ClarifaiAPIExtractor(BatchTransformerMixin, ImageExtractor,
         self.max_concepts = max_concepts
         self.select_concepts = select_concepts
         if select_concepts:
+            select_concepts = listify(select_concepts)
             self.select_concepts = [Concept(concept_name=n) for n in select_concepts]
 
     def _extract(self, stims):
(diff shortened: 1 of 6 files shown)


pliers/extractors/api.py
```diff
@@ -138,31 +152,35 @@ def __init__(self, app_id=None, app_secret=None, model=None,
...
-        else:
-            self.select_classes = ','.join(select_classes)
-
-        self._batch_size = self.tagger.get_info()['max_batch_size']
+        self.api = ClarifaiApp(app_id=app_id, app_secret=app_secret)
+        self.model = self.api.models.get(model)
+        self.min_value = min_value
+        self.max_concepts = max_concepts
+        self.select_concepts = select_concepts
+        if select_concepts:
```
tyarkoni: Should probably check to make sure this is a list (or just `listify` it), otherwise if the user passes a string it'll turn each character into a separate concept.

pliers/extractors/models.py
```diff
@@ -24,6 +24,8 @@ class TensorFlowInceptionV3Extractor(ImageExtractor):
             each image.
      '''
 
+    log_attributes = ('model_dir', 'data_url', 'num_predictions')
```
tyarkoni: should be `_log_attributes` I think

qmac:
Builds are failing cause of an issue with the `indicoio` library, but it should be resolved soon: https://github.com/IndicoDataSolutions/IndicoIo-python/pull/232/files. Going to merge anyways.

PR #183, opened 2017-07-17
Twitter Support

### LINKED ISSUE #131: Add twitter support
It would be pretty neat to add support for feature extraction on tweets. We could do this by, e.g., creating a new `CompoundStim` called `TweetStim` that has slots for `.text`, `.image`, and `.video` (where the first is always populated, and the rest are populated whenever a tweet links to or embeds a media file). We could wrap [python-twitter](https://github.com/bear/python-twitter), so that we could follow the standard pattern for handling API keys. One difference is that we'd probably want something like a `TweetFactory` that takes a set of records retrieved from the twitter API and generates a list of `TweetStim`s. But once that's done, we would in effect have native support for processing a bunch of tweets as if they were TextStim`s, which would be kind of awesome.

qmac:
Got a basic layout for twitter integration for pliers. `TweetStim`s inherit `CompoundStim` such that it can be comprised of text, images, and/or videos. 

Tweets can either by created via the `TweetStimFactory`, which handles the API authentication, or directly by passing a `Status` object from the `python-twitter` library to the `TweetStim` constructor.

Probably could further develop how we get tweets (firehose, other things outside of just status-by-status).

Resolves #131

### DIFF
--- a/pliers/stimuli/api.py
+++ b/pliers/stimuli/api.py
@@ -1,0 +1,45 @@
+import os
+
+from .base import load_stims
+from .compound import CompoundStim
+from .image import ImageStim
+from .text import TextStim
+from .video import VideoStim
+from pliers.transformers import EnvironmentKeyMixin
+
+
+class TweetStim(CompoundStim, EnvironmentKeyMixin):
+
+    _allowed_types = (TextStim, ImageStim, VideoStim)
+    _allow_multiple = True
+    _primary = TextStim
+    _env_keys = ('TWITTER_CONSUMER_KEY', 'TWITTER_CONSUMER_SECRET',
+                 'TWITTER_ACCESS_TOKEN_KEY', 'TWITTER_ACCESS_TOKEN_SECRET')
+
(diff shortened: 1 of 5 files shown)


pliers/stimuli/api.py
```diff
@@ -0,0 +1,78 @@
...
+
+    To get these credentials, visit https://dev.twitter.com/.
+    '''
+
+    _env_keys = ('TWITTER_CONSUMER_KEY', 'TWITTER_CONSUMER_SECRET',
+                 'TWITTER_ACCESS_TOKEN_KEY', 'TWITTER_ACCESS_TOKEN_SECRET')
+
+    def __init__(self, consumer_key=None, consumer_secret=None,
+                 access_token_key=None, access_token_secret=None):
+        import twitter
```
tyarkoni: We may want to use the pattern of importing within a try/except block at the top and setting to `None` if import isn't found, and then checking the `twitter` variable here and raising a comprehensible exception for the user if not available.

PR #185, opened 2017-07-28
Force open images in RGB

### LINKED ISSUE #184: PNG image not being loaded properly (incorrect data dimensions)
Attempt to apply `BrightnessExtractor` (for example) fails on a .png image, but works fine on a .jpg. 
It seems to fail because the following operation is performed on the axis 2, and it is our of range:

            brightness = np.amax(data, 2).mean() / 255.0

It seems as though in the successful case, `stim.data` is 3 dimensional `(768, 1024, 3)` whereas it is only 2 dimensional `(400, 400)`

I imagine some differences between the files is causing the .png to be loaded incorrectly, perhaps because its black and white?

Bad stim:
body01_lcm

adelavega:
To fix #184 we can force RGB when opening images. 

It might be useful to also add BW images to the test suite. Any free images we can use?
Maybe this one? https://en.wikipedia.org/wiki/File:Grayscale_8bits_palette_sample_image.png
Might be overkill though.

### DIFF
--- a/pliers/extractors/image.py
+++ b/pliers/extractors/image.py
@@ -20,6 +20,7 @@ class BrightnessExtractor(ImageExtractor):
 
     def _extract(self, stim):
         data = stim.data
+        assert 0
         brightness = np.amax(data, 2).mean() / 255.0
 
         return ExtractorResult(np.array([[brightness]]), stim, self,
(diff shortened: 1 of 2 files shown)


adelavega:
Tests passed on 2.7. Tests failed on 3.5 but unrealted to my changes (HTTP error)

qmac:
I think we should add tests with that b&w image, to at least the tests for extractors that use 3 channels.

PR #186, opened 2017-08-24
NLP Transformers

### LINKED ISSUE #182: Stemming only works on last word
Doubt this is the intended behavior, but when running the stem filter on a `TextStim` with more than one word, only the last word is stemmed.
Example:
```
>>> stim = TextStim(text='saying')
>>> filt.transform(stim).text
u'say'
>>> stim = TextStim(text='saying something')
>>> filt.transform(stim).text
u'saying someth'
```

qmac:
Expanded text analysis capabilities in pliers, using common NLP techniques.

Added:
-`TextVectorizerExtractor` uses sklearn `Vectorizer`s to extract bag-of-X features from a batch of text documents
-`WordEmbeddingExtractor` uses gensim to extract word embeddings using a user-provided binary file
-`VADERSentimentExtractor` uses nltk's VADER sentiment analysis tool to do sentiment analysis (w/o needing an API)
-`TokenizingFilter` that splits a `TextStim` into several `TextStim`s with one token in each
-`TokenRemovalFilter` that removes specified tokens (e.g. stopwords, punctuation) from a `TextStim`

Resolves #182 by adding a parameter for doing stemming per-token.


pliers/extractors/text.py
```diff
@@ -198,3 +206,95 @@ def _extract(self, stim):
...
+        super(WordEmbeddingExtractor, self).__init__()
+
+    def _extract(self, stim):
+        num_dims = self.wvModel.vector_size
+        if stim.text in self.wvModel:
+            embedding_vector = self.wvModel[stim.text]
+        else:
+            # UNKs will have zeroed-out vectors
+            embedding_vector = np.zeros(num_dims)
+        features = ['embedding_dim%d' % i for i in range(num_dims)]
```
tyarkoni: Maybe make the prefix of the embedding dimension an initializer argument, with the default set to what you hardcoded here (i.e., `"embedding_dim"`)?

pliers/extractors/text.py
```diff
@@ -198,3 +206,95 @@ def _extract(self, stim):
...
+        vectorizer (sklearn Vectorizer): a scikit-learn Vectorizer to extract
+            with. Will use the CountVectorizer by default.
+    '''
+
+    _log_attributes = ('vectorizer',)
+    _batch_size = sys.maxsize
+
+    def __init__(self, vectorizer=None):
+        super(TextVectorizerExtractor, self).__init__()
+        if vectorizer:
```
tyarkoni: We may want to check that `vectorizer` is an instance of an sklearn vectorizer. Alternatively (or in addition), we could let it be a string giving the name of a valid sklearn vectorizer, and initialize it for the user, for convenience.

pliers/filters/text.py
```diff
@@ -49,8 +55,80 @@ def __init__(self, stemmer='porter', *args, **kwargs):
...
+    '''
+
+    _log_attributes = ('tokenizer',)
+
+    def __init__(self, tokenizer=None):
+        self.tokenizer = tokenizer
+        super(TokenizingFilter, self).__init__()
+
+    def _filter(self, stim):
+        if self.tokenizer:
```
tyarkoni: Same suggestion as above: we could validate the type of `tokenizer`, and/or allow a string naming a valid tokenizer in the module.

pliers/filters/text.py
```diff
@@ -49,8 +55,80 @@ def __init__(self, stemmer='porter', *args, **kwargs):
...
+
+    _log_attributes = ('tokens', 'language')
+
+    def __init__(self, tokens=None, language='english'):
+        self.language = language
+        if tokens:
+            self.tokens = set(tokens)
+        else:
+            try:
+                nltk.data.find('corpora/stopwords')
```
tyarkoni: I could be wrong, but I think I wrote a decorator somewhere that allows you to specify what corpora you need, instead of having to specify it inside the function call.

qmac: Yeah I didn't really want to use it since its not explicitly necessary for `_extract` to run, only as a default. Moreover, as a default, it seemed to make more sense to force the download manually.

qmac:
Commit 9d09fa3 provides some tangential fixes, most notably allows `ExtractorResult`s to handle duplicate feature names.

PR #189, opened 2017-09-22
Minor issue fixes

### LINKED ISSUE #118: Consistently handle optional dependency import errors
Often, when importing an optional dependency, a try, catch, except, pass is used. It would be nice if there was a wrapper to warn the user that a dependency is not installed, or just not wrap the import at all and let it crash to make it more obvious what the problem is.

### LINKED ISSUE #135: Add option to silently skip Stims in case where a valid route isn't found
Right now if there's no conversion path from an input `Stim` to what an `Extractor` expects, an exception is raised. This is problematic in cases where we're passing in heterogeneous lists, so we should add the option to silently skip mismatches (or at least, raise a warning and not an exception).

qmac:
Couple random things:
-Updated/standardized optional dependency handling (Resolves #118) by requiring an `attempt_to_import` call at the top and a `@requires_optional_dependency(library_name)` decorator around the `__init__` or `_transform`.
-Added option to silently skip stims (Resolves #135) by adding a `validation` argument to the `.transform` method that accepts values `strict` (will raise TypeError on incorrect validation) `warn` (will just log a warning for incorrect validation) and `loose` (will silently skip and return None)
-Other minor refactoring and clean up


pliers/support/decorators.py
```diff
@@ -19,3 +20,15 @@ def decorated(*args, **kwargs):
...
     return decorated
+
+
+def requires_optional_dependency(dependency):
+    def decorator(func):
+        def wrapper(*args, **kwargs):
+            try:
+                return func(*args, **kwargs)
+            except AttributeError as err:
+                print(err)
```
tyarkoni: Remove this

pliers/support/decorators.py
```diff
@@ -19,3 +20,15 @@ def decorated(*args, **kwargs):
             print(err)
             raise MissingCorpusError()
     return decorated
+
+
+def requires_optional_dependency(dependency):
+    def decorator(func):
+        def wrapper(*args, **kwargs):
+            try:
+                return func(*args, **kwargs)
```
tyarkoni: One danger in wrapping the entire function call is that if it fails with an `AttributeError` that's *not* due to the missing dependency, we'll be giving the user the wrong error message. So I think we need to rework this somehow. A couple of other possibilities:

* We could get rid of the function-level decorator and instead verify that the optional dependency is available at the class level--i.e., we only check once, in `init`. The downside of this is that it will fail even if the user only intends to call functions that *don't* need the offending library. However, this strikes me as unlikely, since in most cases the only interaction with `Transformer`s is through `transform`, so if the object can't be initialized without the requirement, it probably can't be used at all.

* Do it at function level, but make the checks inside the function rather than using a decorator. E.g., something like:

```python

def _extract(self):
    verify_dependencies_exist([sr])
    ...
```

...where the current function call just fails with an Exception if any of the listed dependencies evaluate to `None`. If we want to produce more informative exceptions, we could maintain a global dictionary of optional imports, where the variable name is the key, and the original module name is the value. I actually kind of like this solution, because in principle it would make it quite easy to track cases where *multiple* requirements are missing (and generate an informative message).

qmac: What about just moving the `attempt_to_import` into the function and raise the error in there? Forgot if there was a reason to have the import attempts at the top.

tyarkoni: Main drawback I can see to that it is that it makes it hard to see that the import is happening at a glance, since it's not in the file head. Also, it's slightly more cumbersome code-wise, since we then need to assign the result of the call explicitly, so that instead of having something like `verify_dependencies([lib1, lib2])`, we have the more verbose `lib1, lib2 = attempt_to_import([lib1, lib2])`--possibly multiple times in each module. (We could have the assignment done implicitly inside `attempt_to_import`, but I think that's bad practice, as the assignment is then completely invisible).

On balance I still have a very slight preference towards the pattern I suggested, but if you prefer to do it this way, that's fine with me.

qmac:
@tyarkoni Updated to the pattern you suggested. We talked about making the optional dependency management into a separate package, but I'll get into that later/let's just merge this now. Also had to  fix something (b56287c) that required somewhat of a rewrite of `EnvironmentKeyMixin`.

PR #190, opened 2017-09-24
Librosa Extractors

qmac:
Wrapped audio feature extraction methods from the [librosa package](https://github.com/librosa/librosa).

Including: chroma_stft, chroma_cqt, chroma_cens, melspectrogram, mfcc, rmse, spectral_centroid, spectral_bandwidth, spectral_contrast, spectral_rolloff, poly_features, tonnetz, zero_crossing_rate

1fd3603 had arguments more explicitly listed out for each `Extractor`, but this PR has each Extractor arguments mostly abstracted into the `**librosa_kwargs` parameter.


pliers/extractors/audio.py
```diff
@@ -119,4 +124,259 @@ def _extract(self, stim):
...
+
+    def __init__(self, feature=None, hop_length=512, **librosa_kwargs):
+        verify_dependencies(['librosa'])
+        if feature:
+            self._feature = feature
+        self.hop_length = hop_length
+        self.librosa_kwargs = librosa_kwargs
+        super(LibrosaFeatureExtractor, self).__init__()
+
+    def _get_feature_names(self):
```
tyarkoni: This seems potentially useful to users, so maybe we should drop the leading underscore and expose it publicly. Might actually be a good convention to have for other extractors that have a fixed set of features (I don't think it's worth going back and updating them all, but at least something to keep in mind going forward).

qmac: Sure

PR #196, opened 2017-11-02
Update checker

adelavega:
This is a rough version of an update checker for pliers. 

You provide a path of a csv file, and a graph spec. Optionally some stimuli, but if not provided uses a set of 100 CC0 public domain images I put together. 

If the csv file doesn't exist, it is created and results are saved w/ a timestamp.

If it exists, the script looks for mismatches, appends the new data to the CSV and saves, and reports mismatches and extractors that have changed.

Not sure if you want this in pliers itself. If not, that's fine and I can move it to neuroscout. Might be interested in the stimuli for testing though.


adelavega:
Also, I considering using a sqlite database but I don't really see the advantage over a simple CSV for this purpose.

qmac:
I'm very much a fan of having this in pliers, esp given the feedback we got at KDD

adelavega:
Cool. Do you think its necessary to have video stims to check against? The problem is that you can't apply a video graph to image stimuli. Is there a way to do this? If not, then I can implement a function that would strip a video graph from video-specific converters, and leave only the pure extractors. 

The checker could then return the original graph nodes that require re-extraction (with converters and all)

adelavega:
Great, in the long run, it would also be used full to add video stims to the battery for Video Specific extractors.

IMO the ideal API for this function would be to
In the short term, an idea API for this function would be to give it a graph, and get back nodes of that graph that need to be rerun. If a node begins with a video converter (e.g. `FrameSamplingConverter`) that is not applicable to images (but is not necessary because the input is an image, and the next node requires images), it would ignore this node, and be able to test a "video" node against images.

qmac:
I think we should definitely have video/audio/etc. stims to the battery. Also ok with merging this (once the logic is finished) for now and opening an issue to add those later.

adelavega:
Update checker is pretty much complete, although code is not super optimized. 

Given an input graph such as:

```
{
  "roots": [
    {
      "transformer": "FrameSamplingConverter",
      "parameters": {
        "every": 200
      },
      "children": [
        {
          "transformer": "BrightnessExtractor"
        }
      ]
    },
    {
      "transformer": "VibranceExtractor"
    }
  ]
}
```
`check_updates` will run the Graph on 100 image stimuli. The graph is searched recursively for transformers that are not compatible with the stimuli. In this case `FrameSamplingConverter` is replaced by its child, which does match the input stimuli.

The results are saved in a CSV, with each column representing The result of an extractor for each stimuli, and rows being time stamps. 

Next time this script is run, the new results are compared against the last time point in the csv file. If  there any any mismatches, they are compiled into a list (per stimuli, extractor combination). The set of extractors that has changed is them compiled, and the original Graph is filtered based on those extractors.

If "BrightnessExtractor" changed, the output would be something like:

```
{'changed_extractors': {'BrightnessExtractor'},
 'difference_graph': {'roots': [{'children': [{'transformer': 'BrightnessExtractor'}],
    'parameters': {'every': 200},
    'transformer': 'FrameSamplingConverter'}]},
 'mismatches': ['BrightnessExtractor.brightness.10283671495_c73418908c_z.jpg']}
```

It should be easy then to apply the `difference_graph` to get the new results on your data.

qmac:
Probably want to change `FrameSamplingConverter` to its new location `FrameSamplingFilter`, we recently refactored this code.

adelavega:
@tyarkoni do you wanna take a look at this?

qmac:
Tests would be nice (can just do the `filter` functions if `check_updates` is too heavy)

tyarkoni:
Oh yeah, meant to add that as well. We definitely need tests.

adelavega:
@tyarkoni I simplified `check_updates` to only take Extractors and look for a file in the homedir (unless a new file is passed). We can possibly add the Graph functionality back later, but we'll see if its necessary. 

I added tests and they should be working, with a Dummy Extractor.

pliers/utils/updater.py
```diff
@@ -0,0 +1,80 @@
...
+
+    ## Merge results
+    results = merge_results(results).drop(
+        ['source_file', 'filename', 'history', 'class', 'onset', 'duration'],
+        axis=1, level=0)
+    results['time_extracted'] = datetime.datetime.now()
+
+    # Flatten columns and make into one row
+    results = results.pivot(columns='stim_name', index='time_extracted')
+    results.columns = ['.'.join(col).strip() \
```
qmac: I believe that this should effectively be done if `flatten_columns=True` is passed into `merge_results` (will be separated by underscore instead of period. Along the same lines `metadata=False` can also be passed into `merge_results` and then in L46 you would only need to drop `onset` and `duration`.

qmac:
One minor comment but LGTM

pliers/utils/updater.py
adelavega: I can't remember why I didn't go without because I tried it, but I'll give it another go. Thanks.

adelavega:
Allright looks like this needs to also return the parameters that go along with the Extractor. So in that case, it should probably not take instantiated extractors and instead take names & params, and return the same.

adelavega:
I refactored this to 1) be able to handle filters and converters as well and 2) keep track of parameters, rather than just transformer name.

Now, the input is a list of tuples with transformer name and parameters such as:
    
    [('SomeTransformerName', {'parameter1': 5})]

The updater instantiates these transformers & applies them to the stimuli. In the case of filters or transformers, it extracts the new stimuli and hashes their data (in the case of iterables, it updates this hash with every frame, for example). Results for are saved in CSV.

If anything changes, a mismatch is detected, and the original transformer input (with the parameters specified) is returned as a changed transformer. 

Let me know what you think. I had to change the tests to edit the datastore instead of pass a dummy extract since this function instantiates the transformers for you from the main pliers library.

adelavega:
I also added `__iter__` to `CompoundStim`, to iterate over elements.

adelavega:
@tyarkoni This is ready to merge as far as I'm concerned. Looks like the travis error was downloading related so maybe re-run the tests if you want to doube check, but they pass locally for me.

PR #205, opened 2017-11-14
add Dockerfile, fix error in README

poldrack:
- initial add of Dockerfile
- add env.list to .gitignore to prevent people from committing their credentials file
- fixed incorrect names of Indico API functions in 2nd example in readme


poldrack:
also - I have some documentation about using the Dockerfile, let me know if you would like me to include it in a separate README (and where you would like it to go)

qmac:
I think we should put the Dockerfile and a separate README in a docker/ subdirectory of the repo.

Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+RUN conda create -y -n py35 python=3.5 $CONDA_DEPS
+
+RUN /bin/bash -c "source /miniconda/envs/py35/bin/activate py35 \
+  && pip install --upgrade --ignore-installed setuptools \
+  && pip install python-magic coveralls pytest-cov pygraphviz pysrt xlrd clarifai pytesseract moviepy==0.2.2.13 SpeechRecognition IndicoIo sklearn python-twitter gensim oauth2client google-api-python-client google-compute-engine librosa ipython"
+
+RUN /bin/bash -c 'source /miniconda/envs/py35/bin/activate py35 \
+ && python -c "import imageio; imageio.plugins.ffmpeg.download()"'
+
+RUN git clone https://github.com/tyarkoni/pliers.git
```
qmac: Given that this is within the repo, would it make more sense to just mount and install from local files rather than reclone?

poldrack:
- Dockerfile moved to docker/ 
- README.md added to docker/

Dockerfile
poldrack: sure, I don't have time to work on it now, so do you want to accept the current PR and then add it?

tyarkoni: @qmac I think it's fine to merge as-is and then tweak some more. This doesn't interact with anything else, so there's no risk of breaking any existing functionality. Thanks for the PR, Russ!

qmac: sure, sounds good

Issue #207, opened 2017-11-17
Use audio file metadata to set sampling rate


qmac:
Currently, all `AudioStim`s are set to have a sampling rate of 44100. This is throwing off some of the audio feature extraction and transformation. Instead, we should use the audio's metadata to set this sampling rate. There is not really an elegant way to do this using moviepy, but it should be feasible once https://github.com/Zulko/moviepy/pull/665 is merged and released.

qmac:
Closing #210

PR #208, opened 2017-11-17
Google speech-to-text with onsets

### LINKED ISSUE #117: Fix prefix issue in Google Transformers
Right now the Google API Transformers take a `handle_annotations` argument that can be either `'first'` or `'prefix'` (defaulting to the latter). The problem with this is that when there are multiple results returned, the column names get prepended with a prefix (e.g., `'face1_'`), but when there's only one result, there's no prepending. Consequently, when `merge_results` is called, there are mismatching column names and bad (or at least, confusing) things happen. The way to handle this is probably to split the `handle_annotations` argument into two, where `handle_annotations` now takes values of `'first'` or `'all'`, and we add a new boolean `prefix_first` argument that applies when `handle_annotations='first'`.

qmac:
Moved away from wrapping SpeechRecognition for interfacing with the Google Speech API. This provides the user with more custom control of how they use the API. In particular, we can now provide speech-to-text results that include onsets and durations of words in the transcript.

In addition, resolved some other Google issues:
-The `SafeSearchExtractor` was failing with Python3
-Resolves #117 
-Fixed duration calculation in `VideoFrameCollectionStim`

Uncovered #207 and that needs to be resolved soon.


tyarkoni:
Don't think I'll have time to review this in detail, but I'm sure it's fine, so merge away if ready!

PR #213, opened 2017-12-11
optional reqs - opencv

adelavega:
OpenCV is opencv-python on pypi now


qmac:
Does it still import as the `cv2` module?

adelavega:
Yep. Just tested by uninstall/installing.

qmac:
Sweet

PR #222, opened 2017-12-30
Better config

### LINKED ISSUE #108: Check for and parse config file in user's home folder
Currently all package-wide settings are set in the `config` module. We should probably also check for a .pliersrc file or some such in the user's home dir and update module settings accordingly if one is found, to allow users to set their own sensible defaults.

tyarkoni:
This PR beefs up the config module to bring it more in line with other packages. Instead of directly accessing attributes in `pliers.config` (e.g., `pliers.config.parallelize`), access is now restricted through `config.get_option()` and `config.set_option()` methods.

Additionally, the package now scans in standard locations (user's home directory, and current working directory) for `pliers_config.json` files, which override package-level settings if found. The user can also set the `PLIERS_CONFIG` environment variable with a filename that is handled in the same way.

I'm not 100% satisfied with this--e.g., the get_ and set_ methods are currently inconsistent; it's `.get_option('parallelize')`, but `.set_option(parallelize=True)`). But on the whole I think this is a big improvement.

Closes #108.

### DIFF
--- a/pliers/__init__.py
+++ b/pliers/__init__.py
@@ -1,6 +1,10 @@
+from .config import set_option, get_option
+
 
 __all__ = [
-    'config',
+    'config'
+    'set_option',
+    'get_option',
     'export',
     'google',
     'graph',
--- a/pliers/config.py
+++ b/pliers/config.py
@@ -27,11 +27,15 @@ _default_settings = {
 }
 
 
-def set_option(**kwargs):
+def set_option(key, value):
+    if key not in _settings:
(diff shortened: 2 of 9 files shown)


tyarkoni:
Added you to review, @qmac, but no rush--feel free to get to it when you're back from the break.

qmac:
Whats the opposition to `.set_option(key, value)` combined with `.set_options(**kwargs)`? Other than that LGTM.

tyarkoni:
Good suggestion, will implement it that way.

tyarkoni:
Closing in favor of #228.

Issue #223, opened 2017-12-31
Preserve Extractor results in their original form


tyarkoni:
Currently, we convert the result produced by every `Extractor` to a standardized representation inside the `transform` call--i.e., by constructing an `ExtractorResult` object that already has its values as an array-like object. This makes it easy to combine `ExtractorResults` from different Stims and/or Extractors (e.g., via `merge_results`), but it's problematic in that it means users who expect to get back data in a familiar form have no way of reconstructing the original result.

For example, suppose someone already has existing code to work with results from the Google API, and just wants to use pliers as an easier way to get those results. This kind of thing is probably not going to be a terribly uncommon use case, but we don't support it right now. Instead, we force users to work with new, and likely unfamiliar, representations.

The proposed solution is to let `Extractor` classes pass whatever the wrapped tool would normally return to the user as the first argument (i.e., `data`) to the `ExtractorResult` init. In other words, instead of being something array-like, `data` can be anything at all. This makes it easy for users to, e.g., loop over a list of `ExtractorResult` objects and work with the data in a form native to the underlying tool. BUT to make sure that we can still do all the merging and DataFrame manipulations, we allow `Extractor` classes to optionally implement a `to_df()` class method that take the `data` object (or even an entire `ExtractorResult` instance and return a pandas `DataFrame`. Since `ExtractorResult` instances always store references to the originating `Extractor`, we can then call the Extractor's `.to_df()` method inside `ExtractorResult.to_df()`.

I think this gives us the best of both worlds: retention of the original feature data, regardless of representation, and a way to access the same data in a standardized form. It also gives us a lot of flexibility to take formatting arguments in the `Extractor.to_df()` method--for example, the `handle_annotations` argument in the Google extractors could potentially move to the `to_df()` call, which allows us to keep all data returned from Google, and decide how to format it only when we're actually about to do something with it.

qmac:
Some thoughts:
- Not a huge fan of moving away from standardized output as the default
- We'd have to rework our logging/reproducibility strategies if we go with an `Extractor` level `to_df()`, as any parsing arguments that modify output format would have to be logged/part of identifying stim-->result uniqueness.
- Assuming outputs are still `ExtractorResult` objects, we could just store everything in there as attributes (`raw`, `data`, `values`, or whatever). 
- On the other hand, I do like the idea of abstracting output parsing away from `_extract` methods.

tyarkoni:
[...]

Issue #221, opened 2017-12-29
Rethinking generators


tyarkoni:
In constructing examples for the new docs, I'm coming to the conclusion that our use of generators instead of lists is kind of a UI nightmare. It's pretty much impossible for a naive user (or even me in some cases) to know what kind of iterable they're getting back, which makes inspection of the resulting objects much more cumbersome than it should be. Basically, unless you're planning to go all the way to a pandas `DataFrame` with pliers code (e.g, in a `Graph`), things get hairy pretty quickly.

It's clearly not an optimal solution to just return lists everywhere, since almost anything meaningful one might want to do with, e.g., a large movie, could potentially result in crazy memory use (and this was what prompted me to use generators in the first place). But I think that's at least an easily understandable problem from the user's perspective--i.e., it wouldn't take much for a user to write an outer loop around `VideoFrameStim`s and save the results to file in batches.

Perhaps the right approach is to build batching--and possibly file-writing--functionality into the `Graph`. That way, when users initialize a `Graph`, they just have to specify a batch size, file store, etc. We could potentially use HDF5 to store intermediate results if needed.

In any case, I don't think we should hold up the former change (dropping generators, at least by default) until we have a scheme along the latter lines figured out. We should probably make it a high priority to just use lists for now (and maybe have a config setting that enables them if the user really knows what they're doing).

tyarkoni:
After tinkering a bit, an elegant solution to this problem that requires almost no changes to the code is to keep using generators throughout the core logic (as we currently do), but wrap the to-be-returned iterable (i.e., from the `transform()` call) in a function that checks against a `config.use_generators` boolean. If generators are off (which will be the default now), any generators are recursively converted to lists. Otherwise they're preserved as generators. So basically, we keep using generators internally, but convert them to lists on output, unless the user indicates they want to keep using generators.

Since I've already refactored a bunch of the code as I work up new docs, I'll just integrate this into the forthcoming PR--unless you have any objections, @qmac.

qmac:
This sounds good to me. Is the not knowing if a generator or list is being returned only a product of some of the `BatchTransformerMixin` code (https://github.com/tyarkoni/pliers/blob/master/pliers/transformers.py#L177)? There might be an easy way to fix that but I'd need to look into it more. Either way, the idea of making it a config setting and forcing the returned type sounds good.

tyarkoni:
[...]

PR #228, opened 2018-01-04
Sphinx docs

tyarkoni:
This is a major PR that adds sphinx docs (which can be easily pushed to github.io via the `install_docs.sh` script in the gh-pages branch) as well as a few other major features:

* Sphinx support and revamped docs. The docs are not quite done (most importantly, I'm still working on a quickstart document), but the stuff that's there is basically good to go. Most of the docs came from the current README, but with some fixes and improvements, plus several new sections.

* The config module is much improved and now exposes accessor methods rather than the internal options dict. See #222 for more detail (that PR is now closed, but I've merged it in here).

* Pliers now defaults to returning lists rather than generators whenever iteration is involved. However, users can easily change this via `config.set_option('use_generators', True)`. See #221 for more discussion.

* New `FaceRecognition*Extractor` classes that wrap the face_recognition package. This provides high quality face recognition/location/embedding without requiring any external API.

* Docstrings for all methods/classes/modules that were previously missing them.

* As discussed in #223, all `Extractor`s are now encouraged to store the raw result returned by any underlying extractor in the `.raw` attribute of the returned `ExtractorResult`. They should also ideally implement `to_df` and `to_array` methods that convert `.raw` to pandas DataFrame and numpy array, respectively, as this will likely be the way we handle parsing internally in future. For the time being, this change is completely backward-compatible, but we'll probably need to break the existing API at some point to simplify and streamline things.

* Reorganization of several modules. Both `transformers.py` and `google.py` have now been moved to a top-level `transformers/` directory. The same is true for `utils.py`, which is now in `utils/`, alongside the scikit wrapper and the new update checker.


pliers/config.py
```diff
@@ -1,10 +1,88 @@
...
+
+def from_file(filenames, error_on_missing=True):
+    if isinstance(filenames, string_types):
+        filenames = [filenames]
+    for f in filenames:
+        if exists(f):
+            settings = json.load(open(f))
+            _settings.update(settings)
+        elif error_on_missing:
+            raise ValueError("Config file '%s' does not exist." % f)
```
qmac: Would also be useful to have a method to save the current config settings (default to the home directory standard?). Not high priority

pliers/export.py
```diff
@@ -1,70 +0,0 @@
...
-class Exporter(with_metaclass(ABCMeta)):
-
-    ''' Base exporter class. '''
-
-    @abstractmethod
-    def export(self):
-        pass
-
-
-def to_long_format(df):
```
qmac: I found this method/snippet useful for the Merlin analysis. Could we move it to `utils` or something?

pliers/tests/test_config.py
```diff
@@ -0,0 +1,61 @@
...
+
+    assert opts['n_jobs'] == 200
+    assert opts['log_transformations']
+    assert opts['parallelize']
+
+    reset_options(False)
+
+
+def test_set_option():
+
```
qmac: Extra spacing? Throughout this file

PR #229, opened 2018-01-09
Updating Changelog to reflect 0.2 release

qmac:


qmac:
Not necessary after #234

PR #232, opened 2018-01-22
Better result management

### LINKED ISSUE #227: Package-wide handling of multiple results from a single Stim
It's not uncommon for an `Extractor` to return multiple results for a single `Stim`. For example, there can be multiple faces detected for a single `ImageStim`, which means we then need to somehow represent multiple sets of variables within the same `ExtractorResult`.

This already came up in #77, but I think it's worth reconsidering the solution we settled on in #100--especially in light of discussion in #223 about moving formatting options into the individual `Extractor` classes. In general, I'm not really happy with the current solution of repeating all columns and appending them with suffixes (e.g., 'upper_left_face1', 'upper_left_face2', etc.). I think it often results in huge, unwieldy tables--especially when combined with other results from heterogeneous `Stim` types via `merge_results`.

I think it would be nice to have a package-wide config setting called something like `handle_multiple_results` that provides a variety of options for displaying multiple results. Rather than appending suffixes and repeating columns, I lean towards having the default be to create lists or tuples that contain all of the results. So then instead of having two values like `upper_left_face1=475` and `upper_left_face2=360`, we would would have `upper_left=(475, 360)`. This might create a little more work for users on the parsing side, but it would look much cleaner. We could still provide most/all of the options implemented in the `handle_annotations` argument to the Google extractors (e.g., 'repeat', 'concatenate', 'first', etc.).

One additional benefit of doing it this way is we could pass the result of each `Extractor`'s `to_df` method to the base class's `to_df`, where the reformatting could be done seamlessly, so long as the `Extractor` follows some naming convention. For example, we could mandate that developers have to create a `DataFrame` with multiple repeating sets of columns with numbered suffixes (just like the current handling), and so long as that was respected, end users would get the ability to reformat results for free using any of the above options.

### LINKED ISSUE #230: Default to long-format data when merging ExtractorResults
I don't remember what the original rationale was for returning data from `merge_results` in wide format, but in practice, this becomes kind of unmanageable with heterogeneous extractors because we end up with hundreds of columns and mostly NaN values. I'm thinking of changing the default to long format (with an option to return wide). Any objections, @qmac?

tyarkoni:
This PR makes a number of fairly big (and API-breaking) changes to the way results are managed in `ExtractorResult` objects. Changes include:

* Complete refactoring of the `merge_*` functionality. The `merge_stims` and `merge_features` functions are gone; only `merge_results` remains. The merging code is now much briefer and (I think) easier to understand.
* Ability to specify format as either 'wide' or 'long' in both individual `ExtractorResult.to_df()` calls and `merge_results`. I'm thinking the default will be 'long', per #230, but I could probably be argued out of this. (Note that we're already breaking the `merge_results` API in other ways, so there's not very much incentive to avoid breaking the default here too.)
* A new `extractor_names` argument to `merge_results` that integrates and extends the functionality of the previous `flatten_columns` argument. It's now possible to exercise fairly tight control over how `Extractor` names are displayed in the results (omitted, prepended to feature name, added to a pandas MultiIndex, or stored as a separate column).
* Reworked approach to DataFrame construction for individual `ExtractorResults`. As I suggested in #223 and #227 (and already partly implemented in #228), instead of having `Extractor` classes create array-like objects in their `_transform` methods, the job of converting from data in an Extractor's `.raw` attribute to a DF stored in `.data` is now handled in the base `Extractor`'s `to_df`. Subclasses are expected to implement a `._to_df` method (and optionally, a `_to_array` method) that takes an `ExtractorResult` instance as input, and returns a pandas `DataFrame` as input. This process is optional, however; to ensure backward compatibility, if no `_to_df` method is found, data are handled in much the same way as they were before.
* New expectations for the shape of the DataFrame returned by `to_df`. Going forward, the expectation is that features are always in columns, and multiple records per `Extractor` call are in rows. There are basically two different use cases in which `to_df` returns multiple rows: (1) a single `Stim` has multiple events, each with a different onset and/or duration; and (2) there are multiple objects extracted from the same `Stim`, even though they share onset and duration. An example of the latter occurs when multiple faces are found in a single image. The way the latter should be handled is that the `to_df` call should inject an `'object_id'` column into the DF that explicitly identifies different objects that share onsets and durations (events that reflect a single object but differ in onsets and durations do *not* need a different `object_id`).
* Removed the `to_long_format` I/O helper, as it's redundant with the `format` argument.
* Removed various tests that primarily tested functionality of now-removed methods.
* Made a lot of small changes to various tests to get them to respect the internal changes.

I've updated nearly all of the tests (including ones that need Google and IBM keys, and all the non-API tests), and everything seems to pass. @qmac, if you can prioritize looking at this, I'd appreciate it. Once this is merged, we'll need to go back and adjust some of the existing `Extractor` classes to reflect the new conventions (most notably, the Google extractors that return multiple sets of columns; those should now return just one set of columns, and inject an `'object_id'` column for multiple results). But that's not urgent, since the results are still perfectly usable. Another action item that's even less urgent is to move as much of the DF-creation logic as possible into `_to_df` calls in various extractors.

Closes #230.
Closes #227.


pliers/extractors/base.py
```diff
@@ -92,34 +93,71 @@ def to_df(self, timing=True, metadata=False):
...
+
+        index_cols = ['object_id']
 
         if timing:
             df.insert(0, 'duration', self.durations)
             df.insert(0, 'onset', self.onsets)
+            index_cols.extend(['duration', 'onset'])
+
+        if format == 'long':
+            df = df.melt(index_cols, var_name='feature')
```
qmac: Note: requires pandas >= 0.20

pliers/extractors/base.py
```diff
@@ -82,7 +82,8 @@ def __init__(self, data, stim, extractor, features=None, onsets=None,
             durations = stim.duration
         self.durations = durations if durations is not None else np.nan
 
-    def to_df(self, timing=True, metadata=False):
+    def to_df(self, timing=True, metadata=False, format='wide',
```
qmac: I think the default `format` here should be the same as that of `merge_results`. That being said, still more in favor of defaulting to `wide`.

qmac:
Getting some tests failing:
-Missing attr `merge_stims`: https://github.com/tyarkoni/pliers/blob/better-results/pliers/tests/extractors/test_api_extractors.py#L22
-Same thing: https://github.com/tyarkoni/pliers/blob/better-results/pliers/tests/extractors/test_api_extractors.py#L57
-MultiIndex key error: https://github.com/tyarkoni/pliers/blob/better-results/pliers/tests/extractors/test_api_extractors.py#L112

qmac:
https://github.com/tyarkoni/pliers/pull/232/files#diff-ed370a9901cc2aa08f899c816d2bee46R72 and the next line should be `_to_array`?

pliers/extractors/base.py
```diff
@@ -92,34 +93,71 @@ def to_df(self, timing=True, metadata=False):
...
-                df.columns = features
+            features = self.features
+            if features is None:
+                features = ['feature_%d' % (i + 1)
+                            for i in range(self.data.shape[1])]
+            df = pd.DataFrame(self.data, columns=features)
+
+        # Generally we leave it to Extractors to properly track the number of
+        # objects returned in the result DF, using the 'object_id' column.
+        # But for Extractors that set the _insert_object_id flag to True, we
```
qmac: I don't see any other reference to this `_insert_object_id` attribute

pliers/extractors/base.py
```diff
@@ -132,98 +170,103 @@ def history(self):
...
+        # Set default aggfunc based on column type, otherwise bad things happen
+        if aggfunc is None:
+            aggfunc = 'mean' if is_numeric_dtype(data['value']) else 'first'
+
+        data = data.pivot_table(index=ind_cols, columns='feature',
+                                values='value', aggfunc=aggfunc).reset_index()
+        data[ind_cols] = data[ind_cols].replace('PlAcEholdER', np.nan)
+        data[ind_cols] = data[ind_cols].astype(dict(zip(ind_cols, dtypes)))
+
+    if timing == 'auto' and 'onset' in data.columns:
```
qmac: Would be good to note what this is doing in the argument doc

pliers/extractors/image.py
```diff
@@ -119,11 +119,9 @@ def _extract(self, stim):
...
                                raw=values)
 
-    def to_df(self, result):
-        n_faces = len(result.raw)
-        cols = [self._feature] if n_faces == 1 else \
-            ['%s_%d' % (self._feature, i) for i in range(1, n_faces + 1)]
-        return pd.Series(result.raw, index=cols).to_frame().T
+    def _to_df(self, result):
+        cols = listify(self._feature)
+        return pd.DataFrame([[r] for r in result.raw], columns=cols)
```
qmac: Unrelated to this PR, but should these separate features into the coordinates? I.e. the Google face extractor has a feature for `face_boundingPoly_vertex1_X`, `face_boundingPoly_vertex1_Y`, etc.

pliers/utils/scikit.py
```diff
@@ -42,10 +42,12 @@ def transform(self, stimulus_files):
             result = self.transformer.transform(stimulus_files).to_df()
         else:
             result = self.transformer.transform(stimulus_files, merge=False)
-            result = merge_results(result, flatten_columns=True)
+            result = merge_results(result, format='wide',
+                                   extractor_names=False)
```
qmac: Could probably just put `metadata=False` and `timing=False` instead of the extra code before

pliers/extractors/base.py
qmac: Also not 100% sold on this but at least its configurable I suppose

tyarkoni: Would you prefer to always leave those columns in? My concern is that in a lot of cases, there's no timing information, and then the user always gets back `onset` and `duration` columns that are all `NaN`s, which is both unhelpful and confusing. If we make the default `False`, then we risk users not realizing they can get back timing information--which can be critical. I don't really like the `'auto'` business either, but I'm not sure I see a better alternative...

pliers/extractors/image.py
tyarkoni: I thought about that, but it seems like overkill for the `FaceRecognitionFaceLocationsExtractor`, which just has 4 values, and is not practical for the `FaceRecognitionFaceLandmarksExtractor`, which can have arbitrary numbers of vertices for different features.

pliers/utils/scikit.py
tyarkoni: Point taken, but I think it's helpful to store those extra columns in `metadata_`, so I'll leave it as-is.

tyarkoni:
Ah, the failing tests are all in API-based Extractors I don't have environment variables set for. Can you fix them when you have a chance?

tyarkoni:
And yes, I'd like to include this PR in the 0.2 release. Along with updated README and some more docs, which I'm working on.

pliers/utils/scikit.py
qmac: Oh yea forgot about the `metadata_` attribute 👍

pliers/extractors/image.py
qmac: Agreed on the face landmarks, since there can be an arbitrary number. I believe expanding out the 128 encodings isn't a crazy idea/would be in line with the `WordEmbeddingsExtractor`

qmac:
Yea I can work on getting the tests passing

pliers/extractors/base.py
qmac: Fair enough, we can see how it goes this way.

pliers/extractors/base.py
```diff
@@ -132,98 +170,103 @@ def history(self):
...
-    stims = list(stims.values())
-    return stims[0] if len(stims) == 1 else \
-        ExtractorResult.merge_stims(stims)
+    _timing = True if timing == 'auto' else timing
+
+    if extractor_names is True:
+        extractor_names = 'prepend' if format == 'wide' else 'column'
+
+    dfs = [r.to_df(timing=_timing, metadata=metadata, format='long',
+                   extractor_name=True, add_object_id=True)
```
qmac: Shouldn't we make `add_object_id` a method argument and just pass it through here? On a related note, worth considering whether to do the same "auto" business here when all `object_id`s are 0.

qmac:
When calling `merge_results` with `format=wide`, the index of the resulting DataFrame is called `feature`. Is this intended? It differs from the single `ExtractorResult` method behavior. May be a byproduct of the `pivot_table`.

pliers/extractors/base.py
qmac: Shouldn't we also include the `auto` functionality in the ExtractorResult `to_df` method? Between this and the `object_id` comment below, its kind of tedious to write units tests since its difficult to remember which columns are guaranteed and which aren't between the two methods.

qmac:
To retain consistency with the old API, would be nice to sort resulting DataFrames by onset

pliers/extractors/base.py
tyarkoni: Good idea. I'll add an argument for that, and make the default 'auto'.

pliers/extractors/base.py
qmac: @tyarkoni thoughts on moving the `auto` functionality to the ExtractorResult `to_df` method?

tyarkoni:
Alright, I think we're good to go once the tests pass on travis!

pliers/extractors/base.py
tyarkoni: Oh, sorry. Didn't see this till just now. Yeah, probably. Will update.

tyarkoni: Oh, actually, I just realized that we can't pass the `add_object_id` arg from `merge_results` to `to_df`, because then we risk getting back DFs in `merge_results` (i.e., from different `ExtractorResult` objects) that have different columns (i.e., some with `object_id` and some without). We could of course handle this with another check for the column, but that seems inelegant. So probably I'll just keep the former approach of always requesting the `object_id` column when called from `merge_results`. I'll still add the `auto` option to `to_df` though.

tyarkoni:
Okay, after some more thought, I ended up reverting the defaults for `timing` and `object_id` (which I also renamed from `add_object_id`, for consistency with the `timing`, `metadata`, and `extractor_names` arguments) to `True`. I left `'auto'` in as a valid option though. All tests should be passing (I installed all the optional keys).

qmac:
Ok that definitely sounds good for now

PR #240, opened 2018-01-31
fix: bare installation

mgxd:
currently, if installing into environment without `six`, importing `__version__` within `setup.py` leads to a chain of imports that break installation.

This PR avoids this import and parses the variables within `version.py` directly, as well as allows easier version checking
```
python -c "import pliers; print(pliers.__version__)"
```

### DIFF
--- a/pliers/__init__.py
+++ b/pliers/__init__.py
@@ -1,6 +1,6 @@
 from .config import set_option, get_option, set_options
 from .graph import Graph
-
+from .version import __version__
 
 __all__ = [
     'config'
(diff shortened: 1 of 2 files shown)


tyarkoni:
Thanks! This is an important catch.

Unfortunately the travis tests are now failing because the decryption for the API keys won't work when a PR is coming from a different repo. I want to make sure the tests pass before merging, which will require a fix to the travis setup to avoid this problem in future. I'm busy at the moment but will try to get to it this evening.

tyarkoni:
Thanks!

PR #243, opened 2018-02-01
BF: fixed import - should be pliers.utils.scikit

yarikoptic:


tyarkoni:
Thanks--these examples are all pretty old, so other than the Quickstart, there's a good chance there are other issues.

PR #244, opened 2018-02-01
Temporal cropping

qmac:
Adds two filters that allow for `VideoStim` and `AudioStim` to be trimmed.


qmac:
TODO: use `_optional_input_type`

PR #246, opened 2018-02-02
Minor text transformers updates

### LINKED ISSUE #215: Allow WordEmbeddingExtractor to configure UNK vectors
Currently, any words not found in the `wvModel` of the `WordEmbeddingExtractor` are simply mapped to a zeroed out vector. Could be beneficial for users to be able to configure what values get put in these UNK vectors. 

Potential alternatives include: user-provided values, random Gaussian vectors, nothing at all (which would likely convert to nans if `to_df` is called on the result).

Would be best implemented by adding an argument to the `WordEmbeddingExtractor` initializer.

qmac:
Resolves #215 
Added utility text filter for lower casing
Fixed ordering in `TokenizingFilter`

### DIFF
--- a/pliers/extractors/text.py
+++ b/pliers/extractors/text.py
@@ -242,16 +242,15 @@ class WordEmbeddingExtractor(TextExtractor):
         if stim.text in self.wvModel:
             embedding_vector = self.wvModel[stim.text]
         else:
-            if self.unk_vector is not None:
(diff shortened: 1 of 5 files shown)


tyarkoni:
Looks like there's a merge conflict induced by the last PR; can you resolve it @qmac? Otherwise looks good, so feel free to merge once that's done.

Issue #236, opened 2018-01-31
Avoid exists calls to work with DataLad


adelavega:
`DataLad` `AutomagicIO` monkey patches `open` to allow it to unlock and get files if they are not locally available prior to opening. But, in places where `exists` is called first (e.g. `load_stims`), DataLad won't have a chance to load the file, as `exists` is not monkey patched.

Note they are considering monkey patching `exists`, but I wanted to open this here as it may be an easy fix for my situation to simply try opening files, and catching a `FileNotFoundError`.

qmac:
Linking discussion on https://github.com/datalad/datalad/issues/1924

adelavega:
The update is that Datalad 0.9.x will support `exists` and `isfile`; but only the latter triggers downloading the file. In addition, `libmagic` needs `realpath` once the file is downloaded to get the correct mimetype. I think I can patch this soon as using those two calls should also work with regular files.

qmac:
Closing after #253.

PR #253, opened 2018-02-06
Use isfile and realpath in load_stims to integrate with DataLad

adelavega:
#236 

For DataLad AutoMagicIO to work with unavailable remote files in `load_stims`

Changes `exists` to `isfile` (to trigger download) and then uses `realpath` (which is probably a good thing to do anyways). 

Note, `isfile` must be called on symlink, which is file why I only use `realpath` in `load_file` not earlier. 

Another gotcha is that AutoMagicIO must be activated before `load_stims` is imported.

### DIFF
--- a/pliers/stimuli/base.py
+++ b/pliers/stimuli/base.py
@@ -2,7 +2,7 @@
 
 
 from abc import ABCMeta, abstractmethod
-from os.path import exists, isdir, join, basename
+from os.path import exists, isdir, join, basename, realpath, isfile
 from glob import glob
 from six import with_metaclass, string_types
 from six.moves.urllib.request import urlopen
@@ -128,6 +128,7 @@ def load_stims(source, dtype=None, fail_silently=False):
     }
 
     def load_file(source):
+        source = realpath(source)
         import magic  # requires libmagic, so import here
         mime = magic.from_file(source, mime=True)
         if not isinstance(mime, string_types):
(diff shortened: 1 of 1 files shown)


pliers/stimuli/base.py
```diff
@@ -151,8 +152,9 @@ def load_url(source):
             load_url(s)
         elif isdir(s):
             for f in glob(join(s, '*')):
-                load_file(f)
-        elif exists(s):
+                if isfile(f):
```
tyarkoni: Won't this prevent recursion?

tyarkoni: I think `if isfile(f) or isdir(f)` should do the trick

adelavega: I don't think it was originally meant to support recursion anyways. `load_file` is a separate function, so I think the original code was just looking for flat files in a folder anyways

qmac: I think we should make it a recursive call to `load_stims` (and remove the checks)

adelavega: You mean you expect this function to load all files recursively? I'm not sure that's what I'd necessarily expect. 

And Tal, I think I need to specifically call isfile on each file to trigger Datalad, but I could be wrong. I can check.

qmac: Yea I guess that probably wouldn't be expected, not 100% sure.

tyarkoni: Yeah, agreed. Okay, let's leave it as is.

tyarkoni:
Thanks!

PR #254, opened 2018-02-06
Graph improvements

### LINKED ISSUE #224: to_json and save methods for Graph
The `Graph` class should implement a `to_json` method that returns the json representation of its internal nodes. Then we should wrap that in a `save` method that serializes the graph to a specified filename.

### LINKED ISSUE #225: Improved pygraphviz plots
The PGV graphs generated by `Graph.draw()` are perfectly functional but pretty basic. It would be nice (and very easy) to at minimum add color-coding of Stim and transformer types (extractors vs. filters vs. converters). It might also make sense to indicate (with color, line style, etc.) whether a given transformation was explicitly specified in code, or injected implicitly.

### LINKED ISSUE #249: Simplified Graph API for linear chains of transformers
Currently, it is quite cumbersome to add linear chains to a graph (requires many nested tuples/arrays or JSON objects).

One proposal is to include this functionality as another method of `Graph` such as `add_chain` that adds a list of transformers passed to be executed in sequence.

Another potential approach is to extend the functionality of the `add_nodes` method to allow for adding nodes both vertically (linear chain) or horizontally (branching).

qmac:
Improvements to the Graph API including:
- Allowing graphs to return stims (when calling `.run` with `merge=False`), which is useful for creating `Graph`s for preprocessing pipelines composed of only `Filter`s and `Converter`s
- `to_json` and `save` methods for graph objects for simple serialization and loading
- Ability to add linear chains (sequence of transformers) to a `Graph` using the `add_nodes` method with `mode='vertical'` or the `add_chain` method
- Style updates to the `draw` method, including representing implicit transformations with dotted borders and color coding stimuli and transformers
  - The latter is still a WIP as the color coding needs work aesthetically

Resolves #249
Resolves #225
Resolves #224

### DIFF
--- a/pliers/stimuli/base.py
+++ b/pliers/stimuli/base.py
@@ -212,8 +212,8 @@ class TransformationLog(_trans_log):
     def to_df(self):
         def _append_row(rows, history):
             rows.append(history[:-3])
-            if history[-2]:
-                _append_row(rows, history[-2])
+            if history.parent:
+                _append_row(rows, history.parent)
             return rows
         rows = _append_row([], self)[::-1]
         return pd.DataFrame(rows, columns=self._fields[:-3])
--- a/pliers/stimuli/base.py
+++ b/pliers/stimuli/base.py
@@ -164,7 +164,7 @@ def load_stims(source, dtype=None, fail_silently=False):
     return stims[0]
 
 
-def _log_transformation(source, result, trans=None):
+def _log_transformation(source, result, trans=None, implicit=False):
 
     if result is None or not config.get_option('log_transformations') or \
             (trans is not None and not trans._loggable):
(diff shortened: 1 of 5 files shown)


pliers/stimuli/base.py
```diff
@@ -164,7 +164,7 @@ def load_url(source):
     return stims[0]
 
 
-def _log_transformation(source, result, trans=None):
+def _log_transformation(source, result, trans=None, implicit=False):
```
tyarkoni: Oh, good idea!

pliers/stimuli/base.py
```diff
@@ -208,9 +211,9 @@ def __str__(self):
 
     def to_df(self):
         def _append_row(rows, history):
-            rows.append(history[:-2])
-            if history[-1]:
-                _append_row(rows, history[-1])
+            rows.append(history[:-3])
```
tyarkoni: Since `history` is a namedtuple, we should probably index the attributes by name, which will improve clarity and minimize future edits

qmac: Is there an easy way to do this when its slicing [:-3]?

tyarkoni: you could look up the attribute's index and then use that, but that seems like overkill. For slicing, I think leaving it as-is is fine.

qmac:
This is what the graph coloring looks like now @tyarkoni. Open to changing colors/theming

test

tyarkoni:
Well, it's ugly as shit, but the important distinctions stand out very clearly IMO, and that's what I think we should care about here. Any effort to improve aesthetic appeal is likely to require choosing more similar colors, which would reduce the contrast. So I say merge as-is.

qmac:
Yeah not sure how to win here. Made the colors softer but its still ugly. Allowed the user to pass `color=False` to avoid confusion. Will merge after tests pass.

Issue #258, opened 2018-02-09
Add paid keys for Microsoft APIs to travis


qmac:
I added some temporary keys to travis under `MICROSOFT_FACE_SUBSCRIPTION_KEY`, `MICROSOFT_VISION_SUBSCRIPTION_KEY`, and `MICROSOFT_SUBSCRIPTION_LOCATION`, but these will expire in ~20 days. 

Would be nice to have permanent keys in travis so the tests don't end up failing. Ideally #237 should be addressed as well since this would also incur cost.

Keys can be retrieved from https://azure.microsoft.com/en-us/try/cognitive-services/ and clicking "Get API key" for the Face and Vision APIs (one for each). The location for the keys will also need to be set as a env variable.

tyarkoni:
Cool, will do right now. Do you foresee needing keys for any other services, or just face and vision? If there will be others, I may as well add them now too.

qmac:
Maybe also the TextAnalytics API and the Bing Speech API (speech-to-text). The Emotion, VideoIndexer, and SpeakerID APIs are still in "Preview" so not sure about those.

tyarkoni:
What should `MICROSOFT_SUBSCRIPTION_LOCATION` be set to? I'm not seeing info about that anywhere.

qmac:
It should be at the start of the Endpoint URL. If its the same as mine it should be `westcentralus`

tyarkoni:
Okay, those are all added. The names for the unimplemented ones are `MICROSOFT_BING_SPEECH_SUBSCRIPTION_KEY` and `MICROSOFT_TEXT_ANALYTICS_SUBSCRIPTION_KEY`.

PR #259, opened 2018-02-09
Basic Microsoft Vision APIs

### LINKED ISSUE #216: Add support for Azure Face API
The [Azure Face API](https://docs.microsoft.com/en-us/azure/cognitive-services/face/overview) has some nice features that Google doesn't offer (yet). Most notably, it does face grouping for up to 1,000 faces at a time--which could be really useful for us in a movie context (i.e., to annotate frames by which characters are present).

qmac:
Added support for the Microsoft Cognitive Services Computer Vision APIs - namely the Face API (extracts facial features from images) and the Vision API (extracts many general features from images).

The Face API is capable of annotating the location of faces in images, in addition to landmark, accessories, age, blur, emotion, exposure, facialHair, gender, glasses, hair, headPose, makeup, noise, occlusion, and smile features. The `MicrosoftAPIFaceExtractor` extracts all of these features, or a subset as determined by the `attributes` argument. The `MicrosoftAPIFaceEmotionExtractor` will extract only facial emotion information.

The Vision API is capable of simple object detection similar to Clarifai, Google, Indico, etc. It also contains domain-specific models for image category, image type (clipart, drawing, etc.), colors, and adult content. Each domain-specific feature can be extracted from the various subclasses of `MicrosoftVisionAPIExtractor` or by passing in a list of feature name strings into the `features` argument of `MicrosoftVisionAPIExtractor`.

The Vision API also allows for image-to-text conversion using OCR. This can be done using the `MicrosoftAPITextConverter`.

Ideally, #258 will be addressed soon.
Resolves #216


pliers/converters/microsoft.py
```diff
@@ -0,0 +1,30 @@
...
+
+class MicrosoftAPITextConverter(MicrosoftVisionAPITransformer, ImageToTextConverter):
+
+    ''' Detects text within images using the Microsoft Vision API. '''
+
+    api_method = 'ocr'
+
+    def _convert(self, stim):
+        params = {
+            'language': 'en',
```
tyarkoni: Is `'language'` something we can parameterize and put in the initializer?

pliers/converters/microsoft.py
```diff
@@ -0,0 +1,30 @@
...
+            'detectOrientation': False
+        }
+        response = self._query_api(stim, params)
+
+        text = ''
+        for r in response['regions']:
+            for l in r['lines']:
+                for w in l['words']:
```
tyarkoni: Maybe replace the inner loop with a `join` on a list comprehension for brevity

pliers/extractors/microsoft.py
```diff
@@ -0,0 +1,182 @@
...
+    '''
+
+    api_method = 'analyze'
+    _log_attributes = ('api_version', 'features')
+
+    def __init__(self, features=None, **kwargs):
+        if hasattr(self, '_feature'):
+            self.features = [self._feature]
+        else:
+            self.features = features if features else ['Tags', 'Categories',
```
tyarkoni: Maybe add a check here to make sure any passed values are valid--unless the API ignores invalid ones, in which case it doesn't matter much. Though it's probably still a good idea to raise an exception on invalid values before any API calls are made.

pliers/transformers/base.py
```diff
@@ -215,7 +216,7 @@ def __init__(self, batch_size=None, *args, **kwargs):
             self._batch_size = batch_size
         super(BatchTransformerMixin, self).__init__(*args, **kwargs)
 
-    def _iterate(self, stims, *args, **kwargs):
+    def _iterate(self, stims, validation='strict', *args, **kwargs):
```
tyarkoni: I'm guessing this is just a hook left for later implementation? That's fine with me, just making sure it's not meant to be doing anything right now.

pliers/transformers/microsoft.py
```diff
@@ -0,0 +1,80 @@
...
+
+
+class MicrosoftAPITransformer(Transformer):
+    ''' Base MicrosoftAPITransformer class.
+
+    Args:
+        subscription_key (str): a valid subscription key for Microsoft Cognitive
+            Services. Only needs to be passed the first time the extractor is
+            initialized.
+        location (str): region the subscription key has been registered in.
```
tyarkoni: I think other users might have the same confusion I did, so maybe add a bit more explanation about where this is found (and give an example value or two). Also, does this have to match the user's region? If it all it does is help optimize response latencies, we could probably set a default. I imagine most users won't really care if it takes a few ms longer to get a response.

pliers/transformers/microsoft.py
```diff
@@ -0,0 +1,80 @@
...
+            if self._env_keys not in os.environ:
+                raise ValueError("No Microsoft Cognitive Services credentials "
+                                 "found. A Microsoft Azure cognitive service "
+                                 "account key must be either passed as the "
+                                 "subscription_key argument, or set in the "
+                                 "appropriate environment variable.")
+            subscription_key = os.environ[self._env_keys]
+
+        if location is None:
+            if 'MICROSOFT_SUBSCRIPTION_LOCATION' not in os.environ:
```
tyarkoni: Oh, this addresses my question above. Looks like it's locked to a region. Ignore my above comment (except the part about clarifying what the string is).

pliers/extractors/microsoft.py
qmac: The Extractor will raise an exception when `.transform` is called (the API will return an error code). That ok or should we do it on `__init__`?

tyarkoni: If the error is comprehensible to the user, it's fine as is. If not, we should do it on init.

qmac: The message is `Specified feature type is not valid` so I think that's fine

pliers/transformers/base.py
qmac: That is there to unpack `validation` from `**kwargs`, such that `_transform` isn't called with an unexpected argument. That being said, I'm actually not sure why we have `*args` and `**kwargs` passed into `_transform` since every call should only accept one argument for the stim.

tyarkoni:
Looks good, merge at will!

PR #261, opened 2018-02-16
Miscellaneous updates

### LINKED ISSUE #247: Revise Extractors to follow the new API
Revise `Extractor` classes that have the ability to return multiple undifferentiated rows (i.e., that need an `object_id` assigned). This is mainly the Google `Transformer` hierarchy. While we're at it, we should move all the `handle_annotations` stuff that relates to formatting into `to_df`, and store the raw results in `.raw` rather than formatting them inside the `_transform` call.

### LINKED ISSUE #257: Automatically assign order
Right now the `.order` property of `ExtractorResult`s has to be set by each `Extractor`, which is kind of pointless given that most of the time the order can inferred from other attributes (either onset and duration, or just the order of the list-like data passed to the `ExtractorResult` init). We should probably add some kind of `allow_ordering` flag on `Extractor` classes that, if set to `True`, would allow the `ExtractorResult` init to automatically create the `.order` list internally. That would eliminate the need to worry about order in most `Extractor`s (and the remaining ones could continue to explicitly pass a list of orders as they currently do).

qmac:
Addressing some general improvements:
-Onset, duration, and order are now propagated automatically through on every transformation, if not done in `_transform` (no longer manually required)
-Consolidated the `data` and `raw` arguments back into one mandatory `data` argument
-Implementations for `_to_df` to the appropriate Transformers
-Minor test enhancements for easy wins on test coverage

Resolves #257 - `_propagate_context` method in `.transform`
Resolves #247


tyarkoni:
Let's add something like this:

```python
class ExtractorResult(object):

    def __init__(self, data, **kwargs):
        self.data = data

    @property
    def raw(self):
        return self._data if hasattr(self, 'to_df') else None

    @property
    def data(self):
        return self.to_df()
```

tyarkoni:
Looks good to merge as soon as the tests pass!

qmac:
Whoops just saw that comment, merging.

PR #270, opened 2018-03-02
API transformers updates

### LINKED ISSUE #85: Require explicit permission to run a large set of queries against API extractors?
At the moment, the graph API doesn't do anything to prevent a user from trying to run a full-length movie file through an image extractor, which could result in a very large number of queries (1 per frame) to an API extractor if users aren't careful. It might be a good idea to at minimum issue a warning when a large set of queries (e.g., > 100) to an API `Extractor` is detected, and possibly even require the user to set an explicit flag (e.g., `large_jobs=True`). Alternatively, we could disallow automatic VideoToImageStim conversion in cases where the resulting video frame set is very large.

### LINKED ISSUE #204: Utility function for API testing
To make it easier to determine which APIs are set up correctly, we should add a utility function that makes a test call to every API-based extractor and returns a status code indicating if (a) no API environment variables are set up, (b) environment variables are set up, but the API calls are failing (suggest bad keys), or (c) API variables are set and everything seems to be working. It would be nice to format this as a table similar to the one in the README, so that a user can tell at a glance what's going on.

We already have an `available` property in `EnvironmentKeyMixin` to tell us whether or not keys are available, but we should probably add a (mandatory) method to make a dummy call to the API to see if everything works properly (`test_API_keys`). This will require implementing a dummy call in every API-based extractor (doesn't matter what it returns, if anything, just has to get an "okay" message of some sort from the server in question), but frankly we should probably already be doing that anyway.

### LINKED ISSUE #237: add custom py.test marker to skip/include API-based tests
For the moment, we're not in any danger of running up a serious tab on any of the commercial APIs we wrap (e.g., running all the tests costs about 5 cents a pop in Google Cloud use; other services cost even less). But it would make sense to add a py.test marker for all non-free tests, so that at some point down the line, one can easily skip tests that cost money to run.

### LINKED ISSUE #239: Add example string to API key table in the documentation
For some services, it can be quite confusing to figure out which authentication keys/variables pliers needs. It would probably help to add a column to [this table](http://tyarkoni.github.io/pliers/installation.html#api-keys) in the docs that provides a dummy string with the right format for each variable (e.g., "a1acd84283450ba28ab832022a").

### LINKED ISSUE #271: Rate limits for API transformers
Some commercial APIs (e.g., Microsoft Cognitive Services) have rate limits for free APIs. This means pliers transformations will often result in exceptions even if the user is well under the monthly usage limit for their free tier. To accommodate this, we probably need to add rate-limiting functionality to any transformer making requests.

Now that we have the `APITransformer` class, that's a natural place for this to go. We could probably just keep an internal timer (e.g., `_last_request_time`) that gets reset on every call, and have the `_transform` call go to sleep as needed. We can add an instance variable for the rate limit that by default is set to the class-level default (which will be determined by the service). E.g., for the Microsoft services, we would have `_rate_limit = 3` (20 requests / minute), but the user could pass `rate_limit=10` as an argument if they wanted something else.

### LINKED ISSUE #278: GoogleVisionAPILabelExtractor to_df fails when result empty
Although rare, Google will occasionally return no labels for certain images, like this one (oddly enough):
https://github.com/tyarkoni/pliers/blob/master/pliers/tests/data/image/CC0/4416537561_27d7083571_z.jpg

In that scenario, the raw result will be `[{}]` and the follwing line `to_df` will fail:

    res = {label['description']: label['score'] for label in result._data}

I can fix this one, but I do find it weird this image would return an empty response (I tested others and it seems fine)

qmac:
Variety of improvements and fixes for Transformers that interact with external APIs.
- Added custom marker to unit tests that test APIs that require payments. Such tests can now be skipped by adding `-m "not requires_payment"` to the testing command (e.g. `python3 -m pytest -m "not requires_payment"`)
- `validate_keys` method in every `EnvironmentKeyMixin` class that authenticates the provided API keys
- A new `APITransformer` class with some API-transformer specific validation logic (note: now a significant number of them will also inherit from `BatchTransformerMixin`, and it is important for `APITransformer` to be inherited *first* for the desired behavior. The hierarchy is starting to become muddled, but *functional*.
- Validation in `APITransformer`s that stops jobs of larger than 100 stims or longer than 1 minute from being run if the `config` option `allow_large_jobs` is `False` (default is `True`)
- Validation in `APITransformer` that makes sure `validate_keys` passes before running a transformation
- Updated the API table in `docs/installation.rst` to include `MicrosoftAPITransformers` and another column denoting example values of environment keys to avoid some confusion
- Rate limiting for `APITransformer`s

Resolves #271 
Resolves #239
Resolves #237
Resolves #204 
Resolves #85 
Resolves #278


pliers/utils/base.py
```diff
@@ -117,3 +118,7 @@ def env_keys(cls):
     @classproperty
     def available(cls):
         return all([k in os.environ for k in cls.env_keys])
+
+    @abstractmethod
+    def validate_keys(self):
+        pass
```
qmac: Not implementing this method does not seem to raise an error, as expected with `abstractmethod`s.
I believe that this is a problem since `EnvironmentKeyMixin` is typically the second parent class. Not a big problem and not sure how to address it.

pliers/config.py
```diff
@@ -25,7 +25,8 @@
     'n_jobs': None,
     'parallelize': False,
     'progress_bar': True,
-    'use_generators': False
+    'use_generators': False,
+    'allow_large_jobs': True
```
tyarkoni: We should probably add config keys that define what a large job is here (instead of hardcoding to 100 and 1 minute in the `Transformer` logic

pliers/transformers/api/base.py
```diff
@@ -0,0 +1,38 @@
...
+from pliers.transformers import Transformer
+from pliers.utils import isiterable, EnvironmentKeyMixin
+
+
+class APITransformer(Transformer, EnvironmentKeyMixin):
+
+    def _iterate(self, stims, validation='strict', *args, **kwargs):
+        # Check if we are trying to transform a large number of stimuli
+        if not config.get_option('allow_large_jobs'):
+            stims = list(stims)  # TODO: better way to do this?
```
tyarkoni: Hmm... I don't know that there's any way to check length without consuming a generator. We might have to initialize a counter that gets incremented each time through, and check that way. Otherwise we risk making the `use_generators` config option largely useless.

pliers/transformers/api/base.py
```diff
@@ -0,0 +1,38 @@
...
+        if not config.get_option('allow_large_jobs'):
+            if not isiterable(stim) and stim.duration and stim.duration > 60:
+                raise ValueError("Attempted to run an API transformation "
+                                 "on a stimulus of duration %f, aborting. "
+                                 "To allow this transformation, change "
+                                 "config option 'allow_large_jobs' to "
+                                 "True." % stim.duration)
+
+        if not self.validate_keys():
+            raise ValueError("The provided environment key was invalid or "
```
tyarkoni: It might be good to include the name of the `Transformer` that produced the exception in the message, for more naive users who don't know how to read the stack trace.

pliers/utils/base.py
tyarkoni: Are you setting the metaclass to `ABCMeta`? I think that's required for these decorators to work properly. (Also, since I've gotten burned by this myself recently, note that we'll have to use the `with_metaclass()` function in `six` in order to support this behavior on both Python 2 and 3).

qmac:
@tyarkoni changed to counting the number of transformations. I didn't reset this for each iterable (going away from "large job" definition), but instead made it a per-instance counter.

pliers/transformers/api/base.py
```diff
@@ -2,37 +2,39 @@
...
-                                 "'allow_large_jobs' to True." % len(stims))
-
-        return super(APITransformer, self)._iterate(stims, *args, **kwargs)
+    def __init__(self, **kwargs):
+        self.transformed_stim_count = 0
+        super(APITransformer, self).__init__(**kwargs)
 
     def _transform(self, stim, *args, **kwargs):
         # Check if we are trying to transform a large amount of data
+        self.transformed_stim_count += len(listify(stim))
```
tyarkoni: This will still turn any generator into a list though, which defeats the (memory-saving) point of using generators. (I also think it might consume `stim` if it's a generator, so worth checking to make sure it actually works in that case.)

On further thought, I'm not sure what the best way to handle this is. One possibility is to increment the counter only *after* each transformation is applied. The obvious downside of that is that we risk running 100 transformations and only then raising an exception, which doesn't seem wise.

Alternatively, we could check type here (I think I added an `isgenerator` utility to `utils`) and pre-emptively check for size only if `list` is *not* a generator (and if it is, fall back on applying the transformation anyway). I don't see any way around incrementing the counter in real time for generators, because the way pliers works at the moment, the graph is expanded dynamically, so I don't think there's any way to determine ahead of time how many transformations are going to happen.

I suppose it wouldn't be unreasonable to disable the `allow_large_jobs` option in the case that the user has generators enabled (with a warning or exception if both options are set to `True`). I figure if you've actively decided to use generators, you probably expect to be running large jobs, as generators only help you in the large-memory case anyway. So maybe the path of least resistance is to leave things as-is but issue a warning if needed.

qmac: I will need to triple check this, but I think at this point (in `_transform`), there is no way `stim` is a generator (unless `listify(video)` lists out the frames) since `transform` takes care of iteration. This line of code is necessary for batch transformers, where we already made the concession to listifying prior to transforming.

tyarkoni: Ohhhhh, right. Makes sense. In that case, this works for me. :)

qmac:
Reworked so:
- API-dependent things have a standardized `api_keys` property that returns the values of all required API keys
- Validation is only checked once per `api_keys` via caching
- There is some redundant logic between `TweetStimFactory` and `APITransformer`. Not sure how to solve this without running into multiple inheritance problems (initializing the `validated_keys` attribute)

qmac:
Yay tests are passing......though that required me to manually restart one of the Python2/3 jobs. (when they run simultaneously the rate limit is hit)

qmac:
@tyarkoni thoughts on managing shared logic for API related things? As we put more logic in (e.g. rate limit logic) --> redundancy between `APITransformer` and API-based stim like twitter.

tyarkoni:
I think it would be fine to have a base `APIDependent` mixin that gets subclassed by `APITransformer` and `APIStim`, if that's the kind of thing you're thinking...

qmac:
Yeah thats why I originally put everything in the `EnvironmentKeyMixin` but then we might have to worry about proper `__init__` management (need to actually iron this out)

PR #273, opened 2018-03-20
Fix support for converters to updater

adelavega:
Updater was failing due to a change in the structure of `ExtractorResult` and differences between Convertors and Extractors.

In addition, only the first result may have been checked in extractors, so now I'm hashing all results including those from Extractors

### DIFF
--- a/pliers/utils/updater.py
+++ b/pliers/utils/updater.py
@@ -8,18 +8,14 @@ from os.path import realpath, join, dirname, exists, expanduser
 from pliers.stimuli import load_stims
 from pliers.transformers import get_transformer
 import hashlib
-
+import pickle
 
 def hash_data(data, blocksize=65536):
     """" Hashes list of data, strings or data """
-    if isinstance(data, str):
-        data = [data.encode('utf-8')]
-    if not isinstance(data, list):
-        data = [data]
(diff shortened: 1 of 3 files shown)


pliers/utils/updater.py
```diff
@@ -59,12 +57,11 @@ def check_updates(transformers, datastore=None, stimuli=None):
...
                 try: # Add iterable
-                    res = [res._data for r in res]
+                    res = [getattr(res, '_data', res.data) for r in res]
                 except TypeError:
-                    res = res._data
+                    res = getattr(res, '_data', res.data)
 
-                res = hash_data(res) if isinstance(
-                    trans, (Converter, Filter)) else res[0][0]
+                res = hash_data(res)
```
qmac: Does this work on numpy arrays (e.g. from a LibrosaFeatureExtractor)? JSON (e.g. from a ClarifaiAPIExtractor)?

adelavega:
Okay to fix the issue that the data could be in any format, I chose to pickle all inputs prior to hashing. This should detect any changes, although in combination with hashing the resulting change file will not be human interpretable (e.g. you can't see what changed in the results).

adelavega:
@tyarkoni This look OK to you?

Issue #274, opened 2018-03-23
Clash between feature names and fixed keys (e.g. 'order')


adelavega:
In `ExtractorResult.to_df()`, if feature names are the same as fixed keys (e.g. 'order', 'onset', 'duration'), pandas will raise an error.

In my case, `Clarifai` returned a key named "order" (which was put into a df by: `df = self.extractor._to_df(self, **to_df_kwargs)`). Then, when inserting the column "order" into the df, pandas raised an error: https://github.com/tyarkoni/pliers/blob/772d704cb219235f5daea8a1d0a919bee908d0c9/pliers/extractors/base.py#L153

adelavega:
In chat, the suggestion for the fix was to append `_` to all protected keys

qmac:
Resolved by #281

PR #281, opened 2018-03-29
Append _ to protected names in to_df

adelavega:
Addresses #274 

Waiting on tests. 

Started by only appending `_` to `order`, `duration` and `onset`

Question is if this change should only affect `to_df` or if similar changes should be made throughout pliers. If individual extractors implement `to_df`, do they also need to be changed?

I'm also guessing there will be a lot of downstream effects of changing these column names.

### DIFF
--- a/pliers/extractors/base.py
+++ b/pliers/extractors/base.py
@@ -124,6 +124,11 @@ class ExtractorResult(object):
         durations = np.nan if self.duration is None else self.duration
         orders = np.nan if self.order is None else self.order
 
+        # If any features clash with protected keys, append underscore
+        protected = ['onset', 'order', 'duration', 'extractor', 'stim_name', \
+                     'class', 'filename', 'history', 'source_file']
+        df = df.rename(columns={k: k + '_' for k in protected})
+
         index_cols = []
 
         # Generally we leave it to Extractors to properly track the number of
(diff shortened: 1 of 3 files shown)


qmac:
Why'd you take away the underscore in `extractor` in 5ffe3aa? I get why `feature_` doesn't make sense in long format, but I think it should still be `extractor_` in the `to_df`.

tyarkoni:
@adelavega I think you're inverting the pattern relative to what I was thinking... the event properties `onset`, `duration`, and `order` should *not* have underscores appended in my view. They're the reason these names are reserved. It's the named features `'onset'`, `'duration'`, and `'order'` that get the underscores. Similarly, `extractor` and `object_id` also don't get underscores. We only make changes to returned feature names if they clash, not to the base column names in the DF.

adelavega:
Ahhh shoot. That would have been a lot easier of a change. Will do

tyarkoni:
Is this ready to merge? Looks good to me.

tyarkoni:
Oh, except could you add a test to make sure the renaming is working properly? You can create a fake `ClashingFeatureExtractor` that returns a combination of clashing and non-clashing feature names.

adelavega:
Okay I've added a test. Whats up with coveralls though? Seems to just be failing everything

tyarkoni:
the paid API tests don't run when opening PRs from a fork (to prevent someone malicious from spending all our money). It should be fine once merged--though it's possible some of the API tests will need to be fixed.

PR #283, opened 2018-04-02
Batch caching

### LINKED ISSUE #272: Batch transformations not caching
Since the inputs to a batch transformation are lists of stims instead of stims, memoization is skipped for batch transformers (https://github.com/tyarkoni/pliers/blob/master/pliers/transformers/base.py#L50). 

This creates two problems: 
(1) If a stim is repeated within a batch, it is still run twice
(2) No stim->result transformation is cached. 

We can probably figure out a workaround to (2) but I'm not sure there's a simple solution to (1).

qmac:
Implementation of caching in `BatchTransformerMixin`. This required custom code in `pliers/transformers/base.py` in the `_iterate` method that kept track of repeated elements in a batch or elements already in the cache. 

Also added to some docstrings and the `batch_size` argument everywhere explicitly.

Resolves #272

### DIFF
--- a/pliers/transformers/base.py
+++ b/pliers/transformers/base.py
@@ -236,18 +236,13 @@ class BatchTransformerMixin(Transformer):
         results = []
         for batch in progress_bar_wrapper(batches):
             use_cache = config.get_option('cache_transformers')
-            target_inds = []
+            target_inds = {}
             non_cached = []
-            transformed_keys = set()
             for stim in batch:
                 key = hash((hash(self), hash(stim)))
-                if use_cache and (key in _cache or key in transformed_keys):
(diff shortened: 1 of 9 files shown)


pliers/transformers/base.py
```diff
@@ -235,11 +235,33 @@ def _iterate(self, stims, validation='strict', *args, **kwargs):
         batches = batch_iterable(stims, self._batch_size)
         results = []
         for batch in progress_bar_wrapper(batches):
-            res = self._transform(batch, *args, **kwargs)
+            use_cache = config.get_option('cache_transformers')
```
tyarkoni: Logic all looks good to me, but maybe add a comment or two in this section clarifying what's going on, for posterity.

PR #288, opened 2018-04-19
Video APIs

### LINKED ISSUE #277: Add ability to dynamically set features, onsets, durations, orders in ExtractorResults
Currently any ExtractorResult constructed from `raw` and expecting `to_df` to be called will hold `None` in the `features` `onset` `duration`, and `order` fields after construction.

The latter three are not a particular issue at the moment, since they will be propagated from the `stim` during transformation. However, if `to_df` created multiple feature events, this would be an issue.

Similarly, `features` will always be `None`, and the expected behavior would be that it matches the output of the `data` property.

### LINKED ISSUE #280: Add support for the Google Video Intelligence API
See [here](https://cloud.google.com/video-intelligence/). Looks like it's more or less the same API as the existing Google Cloud APIs, so it shouldn't take much work to add. This should definitely be the highest priority as far as adding new Transformers goes.

qmac:
Provides initial support for the Google Video Intelligence API. 
- `GoogleVideoIntelligenceAPIExtractor` is a general extractor that allows users to specify which features(s) they want to extract, contains main logic for extraction and parsing
- `GoogleVideoAPILabelDetectionExtractor` extracts labels present throughout an entire video, user specified segments, or key frames. This extraction sometimes takes a long time to run, especially if combined with another feature via the parent `GoogleVideoIntelligenceAPIExtractor` class. Use with caution
- `GoogleVideoAPIShotDetectionExtractor` extracts a binary "shot" feature with onsets and durations corresponding to the onsets of shot change detection
- `GoogleVideoAPIExplicitDetectionExtractor` extracts explicit annotations for key frames (ranging from VERY_UNLIKELY to VERY_LIKELY).
- Each transformer is accompanied by `timeout` and `request_rate` parameters that dictate the speed and duration of polling the Google API service for completion

Also expands our support of the Clarifai API by including the ability to extract on videos. This required making the `ClarifaiAPIExtractor` class abstract and refactoring the existing class into a `ClarifaiAPIImageExtractor` and implementing a new `ClarifaiAPIVideoExtractor`.

Resolves #280 
Resolves #277

~~TODO: look for required updates in docs~~


qmac:
@tyarkoni are you ok with this structure of the tests: run the video extraction with a default timeout of 90 seconds
  -If it failed to finish, pass test
  -Otherwise, go through asserts

It will be harder to know when the API has changed and/or is failing, but I can't find a great way to make tests that we know will finish quickly.

pliers/extractors/base.py
```diff
@@ -153,10 +159,12 @@ def to_df(self, timing=True, metadata=False, format='wide',
             df.insert(0, 'onset', onsets)
             df.insert(0, 'duration', durations)
             df.insert(0, 'order', orders)
+            df = df.sort_values('onset').reset_index(drop=True)
             index_cols.extend(['onset', 'order', 'duration'])
 
         if format == 'long':
             df = df.melt(index_cols, var_name='feature')
+            df = df.dropna(subset=['value'])
```
qmac: Drops any rows with NaN values in long format

pliers/extractors/base.py
```diff
@@ -120,9 +120,15 @@ def to_df(self, timing=True, metadata=False, format='wide',
...
-        orders = np.nan if self.order is None else self.order
+        if hasattr(self, '_onsets'):
+            onsets = np.array(self._onsets)
+            onsets += 0.0 if self.onset is None else self.onset
+        else:
+            onsets = np.nan if self.onset is None else self.onset
+        durations = getattr(self, '_durations', self.duration)
+        durations = np.nan if durations is None else durations
+        orders = getattr(self, '_orders', self.order)
+        orders = np.nan if orders is None else orders
```
qmac: Manually checks to see if the Extractor has set new custom timing attributes (prepended with _) during a call to `_to_df`

pliers/extractors/api/google.py
```diff
@@ -132,3 +138,233 @@ def _to_df(self, result):
...
+        msg = "Beginning video extraction with a timeout of %fs. Even for "\
+              "small videos, full extraction may take awhile." % self.timeout
+        logging.warning(msg)
+
+        operation_start = time.time()
+        response = self._query_operations(operation['name'])
+        while 'done' not in response and \
+              (time.time() - operation_start) < self.timeout:
+            response = self._query_operations(operation['name'])
+            time.sleep(self.request_rate)
```
qmac: Polls the Google Video Intelligence Operation for completion every `self.request_rate` seconds

qmac:
@tyarkoni ready for final review. I commented on the major issues/things in the PR since it's pretty big

pliers/converters/api/ibm.py
```diff
@@ -65,7 +65,7 @@ def check_valid_keys(self):
             self._send_request(request)
             return True
         except Exception as e:
-            if 'Unauthorized' in str(e):
+            if 'Not Authorized' in str(e):
```
qmac: Fixes current test failures on `master`

PR #296, opened 2018-05-03
Fix missing RGB

### LINKED ISSUE #295: GooglePropertyExtractor - VisionAPI not returning full RGB information
This look more like a bug with GoogleVisionAPI, but for some images (er rather just one), only "red" and "green" are returned in the response, but no blue, leading to a KeyError. I'm going to investigate if its because blue = 0, or because of a random fluke. It would be good to make the extractor robust to these possible responses.

    color = {'color': {'red': 68, 'green': 3}, 'score': 0.007824038, 'pixelFraction': 0.014705882}

https://github.com/tyarkoni/pliers/blob/a7ea0d2c69de92bf74388295de0a18311bf0a294/pliers/extractors/api/google.py#L112

adelavega:
Fixes #295

### DIFF
--- a/pliers/extractors/api/google.py
+++ b/pliers/extractors/api/google.py
@@ -109,7 +109,9 @@ class GoogleVisionAPIPropertyExtractor(GoogleVisionAPIExtractor):
         data_dict = {}
         for color in colors:
             rgb = color['color']
-            data_dict[(rgb.get('red',0), rgb.get('green', 0), rgb.get('blue', 0))] = color['score']
+            key = [rgb.get('red', 0), rgb.get('green', 0), rgb.get('blue', 0)]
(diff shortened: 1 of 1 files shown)


qmac:
Potentially would not have occurred if we were wrapping `google-cloud-python` instead of directly requesting. But LGTM.

adelavega:
Yes, looks like they fixed that in their python module, but not the API itself.

adelavega:
Before this gets merged, I just found a related bug that should be addressed: the "feature_name" in `to_df` for this extractor is the R value, not the RGB combined.

adelavega:
Fixed. @tyarkoni let me know if this looks reasonable.

PR #299, opened 2018-05-11
Forced alignment

### LINKED ISSUE #290: PunctuationRemovalFilter only removes punctuation that are tokens
Most notably, apostrophes linger.
Simple solution: don't inherit from `TokenRemovalFilter`, just use a simple regex replace.

qmac:
Provides basic forced alignment via the python package [aeneas](https://github.com/readbeyond/aeneas). Forced alignment is performed via a `Filter` operating on a `TranscribedAudioCompoundStim` that contains an audio stim and its corresponding transcript. Alignment is done by providing onsets and duration to each component `TextStim` in the `ComplexTextStim` transcript. An example script performing word-level alignment is added under `examples/`. Note that results are barely mediocre, particularly since `aeneas` was not built to perform word-level alignment.

Currently exploring other packages with more sophisticated setup that perform forced alignment better, especially at the word granularity.

This contains some additional maintenance/bug fixes that could be cherry-picked if we decide not to merge and/or want those fixes in master sooner. (Resolves #290)

Travis seems to dislike installing `aeneas` via pip, not sure what to do about that.


qmac:
Closing in light of the bug fixes being in #301 and the results being fairly poor

PR #301, opened 2018-05-15
Various bug fixes and minor cleanups

### LINKED ISSUE #290: PunctuationRemovalFilter only removes punctuation that are tokens

### LINKED ISSUE #294: VideoFrameStim initialized from data will reload video
Haven't explicitly tested this, but looking at the code it appears the constructor for `VideoFrameStim`, when give empty `data` will call the `ImageStim` constructor with the `VideoStim` filename, causing the data to be reloaded (may even error, but worst case an expensive operation). The `data` is then immediately overwritten with the correct frame, but we should avoid the reloading.

Fortunately, I don't think there's anywhere in the codebase where this actually happens.

### LINKED ISSUE #298: get_frame using onset on a sampled VideoFrameCollection IndexError
https://github.com/tyarkoni/pliers/blob/master/pliers/stimuli/video.py#L102 finds the frame number, not the index.

qmac:
A couple patches to the mechanics of `VideoStim` and `VideoFrameCollectionStim`.
-Changed the `get_frame` method to only accept an `onset` argument on `VideoStim`s, since this becomes ambiguous for `VideoFrameCollectionStim`s anyways
-Updates to make the sampling more accurate (rounded at the appropriate moment) in the `FrameSamplingFilter`. Made it possible to perform upsampling, but took away the ability to sample on an already-sampled `VideoFrameCollectionStim`

Resolves #294
Resolves #298

Cherry picked some commits from #299.
-Resolves #290. Removes punctuation that count as tokens
-Small update to the `speech_contexts` argument for the `GoogleSpeechAPIConverter`
-Small fix for audio sampling rate if initialized from an existing clip
-Added functionality to `ComplexTextStim` to enable output to .srt files

### DIFF
--- a/pliers/filters/text.py
+++ b/pliers/filters/text.py
@@ -139,8 +139,11 @@ class PunctuationRemovalFilter(TokenRemovalFilter):
 
     ''' Removes punctuation from a TextStim. '''
 
-    def __init__(self, tokens=string.punctuation):
-        super(PunctuationRemovalFilter, self).__init__(tokens)
+    def _filter(self, stim):
+        text = stim.text
+        for c in string.punctuation:
+            text = text.replace(c, '')
+        return TextStim(stim.filename, text)
 
 
 class LowerCasingFilter(TextFilter):
--- a/pliers/filters/text.py
+++ b/pliers/filters/text.py
@@ -2,6 +2,7 @@
 
 import nltk
 import string
+import re
 
 from six import string_types
 from nltk import stem
@@ -135,14 +136,13 @@ class TokenRemovalFilter(TextFilter):
         return TextStim(stim.filename, text)
 
 
-class PunctuationRemovalFilter(TokenRemovalFilter):
+class PunctuationRemovalFilter(TextFilter):
 
     ''' Removes punctuation from a TextStim. '''
 
     def _filter(self, stim):
-        text = stim.text
-        for c in string.punctuation:
-            text = text.replace(c, '')
+        pattern = '[%s]' % re.escape(string.punctuation)
+        text = re.sub(pattern, '', stim.text)
         return TextStim(stim.filename, text)
 
 
--- a/pliers/filters/video.py
+++ b/pliers/filters/video.py
@@ -1,9 +1,13 @@
 ''' Filters that operate on TextStim inputs. '''
 
 from pliers.stimuli.video import VideoStim, VideoFrameCollectionStim
+from pliers.utils import attempt_to_import, verify_dependencies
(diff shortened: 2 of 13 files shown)


pliers/filters/text.py
```diff
@@ -139,8 +139,11 @@ class PunctuationRemovalFilter(TokenRemovalFilter):
 
     ''' Removes punctuation from a TextStim. '''
 
-    def __init__(self, tokens=string.punctuation):
-        super(PunctuationRemovalFilter, self).__init__(tokens)
+    def _filter(self, stim):
+        text = stim.text
+        for c in string.punctuation:
```
tyarkoni: Could replace this with `text = text.translate(None, string.punctuation)`, which should be much faster.

pliers/filters/video.py
```diff
@@ -1,17 +1,26 @@
...
+
 
 class VideoFilter(Filter):
 
     ''' Base class for all VideoFilters. '''
 
+    _input_type = VideoStim
+
+
+class VideoFrameFilter(Filter):
```
tyarkoni: I think we can probably drop this if the only thing it's used for is to set the `_input_type`... that could just be done in each subclass. Unless you think it will make sense to add functionality here down the line, or there's somewhere else it's important.

pliers/filters/video.py
```diff
@@ -36,28 +45,33 @@ def __init__(self, every=None, hertz=None, top_n=None):
...
-            new_idx = range(int(video.fps * video.clip.duration))[::self.every]
+            new_idx = range(video.n_frames)[::self.every]
         elif self.hertz is not None:
-            interval = int(video.fps / self.hertz)
-            new_idx = range(int(video.fps * video.clip.duration))[::interval]
+            interval = video.fps / float(self.hertz)
+            new_idx = np.arange(0, video.n_frames, interval).astype(int)
+            new_idx = list(new_idx)
         elif self.top_n is not None:
             import cv2
```
tyarkoni: Should probably use the import checking util here to make sure `cv2` is available.

pliers/filters/text.py
qmac: Doesn't seem to be a clean solution this way that supports Python 2 and 3?

tyarkoni: Ah, didn't realize this wasn't in 3. We could import `maketrans` from different places and construct the argument to `translate` that way, but it doesn't seem worth it. I think at minimum we should be able to use `re.gsub` though, which should still be much faster.

PR #308, opened 2018-07-11
Add os.path and pliers.tests.utils imports

rogilmore:
I found that I needed these to make the Quick Start demo work.


qmac:
Thanks for pointing this out and fixing it! It may be worth to put it in a separate section at the top of the doc, like it is in the example notebook, as it is used throughout the instructions.

tyarkoni:
Thanks Rick!

PR #312, opened 2018-09-18
Set `unknown` fps to 44100

### LINKED ISSUE #311: Loading videos: Audio FPS can be unknown
When loading a video, an error occurs due to this line:

https://github.com/tyarkoni/pliers/blob/cfd43d4143fd8a3b148cf12c23996128b671d428/pliers/stimuli/video.py#L82

audio_fps is return as `unknown`,  and that value later causes `moviepy` to throw an error, because it is expecting a float, not string.

seems as though ffmpeg is returning `unknown` for some reason, and previously this wasn't a problem because the `audio_fps` was previously not passed down in the same manner. 

I think the easy solution here is to treat `unknown` as None, and set to the default value of `44100`

adelavega:
Closes #311


tyarkoni:
Thanks!

PR #316, opened 2018-12-05
Changed location of words list assignment for GoogleSpeechAPIConverter

amekhanik:
In using the GoogleSpeechAPIConverter within pliers I noticed a mistake in the location where the list containing all the transcribed words is assigned. The list is re-assigned with every new response from the API and so whatever was in all of the previous responses would get overwritten and only the last speech block will be returned as the final output.

I just quickly moved it outside the for loop so if it's transcribing a clip with multiple audio blocks, everything would get returned.

### DIFF
--- a/pliers/converters/api/google.py
+++ b/pliers/converters/api/google.py
@@ -90,10 +90,10 @@ class GoogleSpeechAPIConverter(GoogleAPITransformer, AudioToTextConverter):
             raise Exception(response['error']['message'])
 
         offset = 0.0 if stim.onset is None else stim.onset
+	words = []
         if 'results' in response:
             for result in response['results']:
                 transcription = result['alternatives'][0]
-                words = []
                 for w in transcription['words']:
                     onset = float(w['startTime'][:-1])
                     duration = float(w['endTime'][:-1]) - onset
--- a/pliers/converters/api/google.py
+++ b/pliers/converters/api/google.py
@@ -90,7 +90,7 @@ class GoogleSpeechAPIConverter(GoogleAPITransformer, AudioToTextConverter):
             raise Exception(response['error']['message'])
 
         offset = 0.0 if stim.onset is None else stim.onset
-	words = []
+        words = []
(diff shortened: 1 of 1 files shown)


qmac:
Great catch @amekhanik, will merge soon! Thanks!

amekhanik:
Hey @qmac quick thought, would it make sense to separate the output from each audio block? So each response has its own list of words, all housed within a total transcription list? This may make sentiment analysis performance better, instead of amalgamating all the text for a 45 second clip, for example, where multiple sentiments may get meshed together.

Let me know what you think, I'd be glad to make the quick change

qmac:
Hmmm, @amekhanik which of these scenarios are you describing for the output of this Converter?
1) Output a `ComplexTextStim` where every component `TextStim` is a space-delimited text segment of the audio block
    - This would be analogous to outputting ["here is the first block", "here is the second block"]
2) Output a list of `ComplexTextStim` where every component `ComplexTextStim` corresponds to an audio block, whose component `TextStim` are single words
    - This would be analogous to outputting [["here", "is", "the", "first", "block"], ["here", "is", "the", "second", "block"]]
3) Something else?

With the current changes from your PR, this outputs a single `ComplexTextStim` of component `TextStim`s that are words, analogous to ["here", "is", "the", "first", "block", "here", "is", "the", "second", "block"].

This discussion has come up before, and we typically try to limit the amount of tokenization/ detokenization imposed by the converter, as this is a messy endeavor. 

Do you know if the result blocks from Google are representative of any meaningful split (speaker change, etc.)? My initial guess was no, which is why the code was written as is. It also helps standardize the output with the other speech-to-text converters.

amekhanik:
Yeah I realized a little after posting that that a ComplexTextStim can't iterate through nested lists as it expects every element to be a TextStim so it can use the onset. And I had in mind the 2nd scenario. Yeah the result blocks split from Google seem to just represent a long enough period of no transcribable speech

PR #319, opened 2019-02-04
initial review for changes to text.py

jayeetaroy:


pliers/extractors/text.py
```diff
@@ -13,9 +14,16 @@
 import pandas as pd
 import nltk
 from nltk.sentiment.vader import SentimentIntensityAnalyzer
-import sys
+
+import re
+import spacy
```
tyarkoni: Instead of importing directly, the pattern we use is to assign the import using the `attempt_to_import` helper; see the `sklearn_text` example on [this line](https://github.com/tyarkoni/pliers/pull/319/files#diff-7f7f2c63289715b624c2a87196693f96R29). You would do something similar for spacy.

pliers/extractors/text.py
```diff
@@ -155,6 +163,20 @@ def _extract(self, stim):
...
 
+class LowerCaseSpecialCharStripExtractor(TextExtractor):
+
+    ''' Strips the text of the special characters and converts to lower case. '''
+
+    VERSION = '1.0'
+
+    def _extract(self, stim):
+        
+        string=re.sub('[^ a-zA-Z0-9]', '', stim)
```
tyarkoni: This needs to be `stim.text`; the input `stim` is a `TextStim` instance, not a string. The text is stored in its `.text` attribute.

pliers/extractors/text.py
```diff
@@ -313,3 +335,38 @@ def _extract(self, stim):
         features = ['sentiment_' + k for k in scores.keys()]
         return ExtractorResult([list(scores.values())], stim, self,
                                features=features)
+
+class SpaCyExtractor(TextExtractor):
+    
+    def __init__(self, model='en_core_web_sm', features=None):
+        #verify_dependencies(['spacy'])
```
tyarkoni: This is the right approach; you can uncomment this line and `spacy` will be available in the namespace (assuming you've imported with `attempt_to_import` as mentioned above).

pliers/extractors/text.py
```diff
@@ -313,3 +335,38 @@ def _extract(self, stim):
...
+        try: 
+            nlp = spacy.load('en_core_web_sm')
+        except ImportError:
```
tyarkoni: Rather than raising an exception if `spacy.load`  fails, you could first try downloading it via `spacy_download`. One way to do this is to have a second, nested try/except block. I.e., the logic is: first try to load the desired language model. If it fails, try to download it. If *that* fails, only then do we raise an `ImportError`.

pliers/extractors/text.py
```diff
@@ -313,3 +335,38 @@ def _extract(self, stim):
         features = ['sentiment_' + k for k in scores.keys()]
         return ExtractorResult([list(scores.values())], stim, self,
                                features=features)
+
+class SpaCyExtractor(TextExtractor):
+    
+    def __init__(self, model='en_core_web_sm', features=None):
+        #verify_dependencies(['spacy'])
+        try: 
+            nlp = spacy.load('en_core_web_sm')
```
tyarkoni: I imagine this was probably just a first step, but don't forget to replace `'en_core_web_sm'` with `self.model`, as we want to load whatever model the user wants to use.

pliers/extractors/text.py
```diff
@@ -313,3 +335,38 @@ def _extract(self, stim):
         features = ['sentiment_' + k for k in scores.keys()]
         return ExtractorResult([list(scores.values())], stim, self,
                                features=features)
+
+class SpaCyExtractor(TextExtractor):
+    
+    def __init__(self, model='en_core_web_sm', features=None):
```
tyarkoni: Don't forget to store the passed arguments internally in the instance (i.e., to assign `self.model = model`, etc.) That way you can access them later from the `_transform` method.

pliers/extractors/text.py
```diff
@@ -313,3 +335,38 @@ def _extract(self, stim):
...
+            nlp = spacy.load('en_core_web_sm')
+        except ImportError:
+            logging.warning("Spacy Models ('en_core_web_sm') not found. Downloading and installing")
+            spacy_download("en_core_web_sm-2.0.0/en_core_web_sm-2.0.0.tar.gz")
+           
+            
+            
+    def _extract(self, stim):
+        #make lowercase, remove puntuations:
+        stim=LowerCaseSpecialCharStripExtractor._extract(stim)
```
tyarkoni: Let's drop this part. It's a good thought, but for now let's leave out any preprocessing. (If the user wants to do some, they can explicitly include a preprocessing transformer in their chain.)

pliers/extractors/text.py
```diff
@@ -313,3 +335,38 @@ def _extract(self, stim):
...
+        features=[[]]
```
tyarkoni: As we discussed, you were on the right track: the data returned is a 2D array (though I believe you can also pass it to `ExtractorResult` as a list of lists; either should work). Each column is a feature value, and each row is (in our case) a word in the sequence. I'll add more comments once you revise this chunk of the code to be more general.

tyarkoni:
Note also that the automated tests are currently failing because spacy isn't being installed on Travis-CI. You can fix this by updating the `travis.yml` config file to make sure that `spacy` is pip installed alongside the other dependencies.

pliers/extractors/text.py
```diff
@@ -13,14 +14,20 @@
 import pandas as pd
 import nltk
 from nltk.sentiment.vader import SentimentIntensityAnalyzer
-import sys
+import os
+import re
+from spacy.cli.download import download as spacy_download
```
tyarkoni: I think you can drop this line; `spacy` should already be in the namespace thanks to the `attempt_to_import` line below, which implicitly imports it (if it's installed)

pliers/extractors/text.py
```diff
@@ -313,3 +334,52 @@ def _extract(self, stim):
...
+    def __init__(self, model='en_core_web_sm', features=None):
+        verify_dependencies(['spacy'])
+        try: 
+            print(model)
+            model_na=model[:2]
+            self.model = spacy.load(model_na)
+            self.features= features
+            print('loaded model: ',self.model )
+        except (ImportError, IOError, OSError) as e:
+            logging.error("Spacy Models ('"+model+"') not found. Downloading and installing")
```
tyarkoni: This works, but it's conventional and more maintainable to use string formatting for this kind of thing—i.e., something like `"Spacy Model '{}' not found. Downloading and installing".format(model)`

pliers/extractors/text.py
```diff
@@ -313,3 +334,52 @@ def _extract(self, stim):
...
                                features=features)
+
+class SpaCyExtractor(TextExtractor):
+
+    
+    def __init__(self, model='en_core_web_sm', features=None):
+        verify_dependencies(['spacy'])
+        try: 
+            print(model)
+            model_na=model[:2]
```
tyarkoni: This is purely a stylistic thing that has no functional implications, but we try to adhere to the PEP8 style guide. Per PEP8, operators should have whitespace on both sides (i.e., `a = 2`, not `a=2`). This may seem pedantic, but it's good to get into the habit of following the conventions, because it makes code more readable (and most other Python codebases also follow PEP8, so it will benefit you later). Depending on your IDE/text editor, you can probably install a plugin that auto-formats the code for you.

pliers/extractors/text.py
```diff
@@ -313,3 +334,52 @@ def _extract(self, stim):
...
+           
+            
+            
+    def _extract(self, stim):
+        
+        inputTxt=self.model(stim.text)
+        features_list=[]
+        #taking the sentiment value for example 
+    
+        if self.features==None:
```
tyarkoni: When checking against `None`, the Python idiom is to use `is` rather than `==`.

pliers/extractors/text.py
```diff
@@ -313,3 +334,52 @@ def _extract(self, stim):
...
+            #spacy_download.download_model("en_core_web_sm-2.0.0/en_core_web_sm-2.0.0.tar.gz")
+            os.system('python -m spacy download '+model)
+            
+        super(SpaCyExtractor, self).__init__()
+           
+            
+            
+    def _extract(self, stim):
+        
+        inputTxt=self.model(stim.text)
```
tyarkoni: Per Python naming conventions, variable names should use snake case (i.e., `input_txt`).

pliers/extractors/text.py
```diff
@@ -313,3 +334,52 @@ def _extract(self, stim):
...
+            features=['raw_text', 'text','lemma', 'pos', 'tag', 'syntactic_dependancy', 'shape', 'alpha', 'stop' ]
+        
+        for token in inputTxt:
+            arr=[]
+            arr.extend((token, token.text, token.lemma_, token.pos_, token.tag_, token.dep_,
+                  token.shape_, token.is_alpha, token.is_stop))
+            features_list.append(arr)
+            
+        #print(features_list)  
+        return ExtractorResult(features_list, stim,
```
tyarkoni: Let's also pass `order` in—that can just be a sequential list of integers in this case.

pliers/extractors/text.py
```diff
@@ -313,3 +334,52 @@ def _extract(self, stim):
...
+        inputTxt=self.model(stim.text)
+        features_list=[]
+        #taking the sentiment value for example 
+    
+        if self.features==None:
+            features=['raw_text', 'text','lemma', 'pos', 'tag', 'syntactic_dependancy', 'shape', 'alpha', 'stop' ]
+        
+        for token in inputTxt:
+            arr=[]
+            arr.extend((token, token.text, token.lemma_, token.pos_, token.tag_, token.dep_,
```
tyarkoni: This will handle the case where the user didn't pass in features themselves, but if `features is not None` at initialization, the user will still get back all of the features. You'll need to loop over features and only add them if they're in `features`.

pliers/extractors/text.py
```diff
@@ -342,44 +341,44 @@ def __init__(self, model='en_core_web_sm', features=None):
         verify_dependencies(['spacy'])
         try:
```
tyarkoni: If the `try` block fails, `self.model` and `self.features` won't exist, so extraction will fail when the user calls `transform()`, even if the download process succeeded.

pliers/extractors/text.py
tyarkoni: Note: `input_txt`, not `input_Txt`.

pliers/extractors/text.py
```diff
@@ -313,3 +332,114 @@ def _extract(self, stim):
...
+     tag_ : Fine-grained part-of-speech.(unicode)
+     dep_ : Syntactic dependency relation.(unicode)
+     shape_ : Transform of the tokens's string, to show orthographic features. For example, "Xxxx" or "dd".(unicode)
+     is_alpha : Does the token consist of alphabetic characters? (bool)
+     is_stop : Is the token part of a "stop list"? (bool)
+     is_punct : Is the token punctuation?(bool)
+     sentiment : A scalar value indicating the positivity or negativity of the token.(float)
+     is_ascii : Does the token consist of ASCII characters?(bool)
+     is_digit : Does the token consist of digits?(bool)
+    '''
```
tyarkoni: The docstring should also have a separate `Args:` section that describes what the arguments (`model` and `features` are). See other `Extractors` for examples.

pliers/extractors/text.py
```diff
@@ -313,3 +332,114 @@ def _extract(self, stim):
...
+            self.model = spacy.load(model)
+
+            print('loaded model: ', self.model)
+        except (ImportError, IOError, OSError) as e:
+            logging.error(
+                "Spacy Models ('{ "+model+" }') not found. Downloading and installing")
+
+            os.system('python -m spacy download '+model)
+            model_na = model[:2]
+            self.model = spacy.load(model_na)
```
tyarkoni: I'm assuming the reason you're taking just the language part of the model name is to try and load a related model if the specific one the user asked for can't be found. But we probably don't want to do that in this case, because then the user won't know that we loaded a different model.

I.e., suppose they pass `model='en_new_fancy_model'`. That doesn't actually exist, but because spacy will successfully load *something* if we try to load just `'en'`, we're now in a position where we no longer know exactly which model was loaded. We could in theory track this internally, but it's probably easier and safer to just fail. So let's change this to just `load(model)`, and if it fails, the user will know that they're passing an invalid model name.

pliers/extractors/text.py
```diff
@@ -313,3 +332,114 @@ def _extract(self, stim):
...
+        super(SpaCyTokenExtractor, self).__init__()
+
+    def _extract(self, stim):
+
+         input_txt = self.model(stim.text)
+         features_list = []
+
+         order_list = list(range(1, len(input_txt)+1))
+
+         if self.features is None:
```
tyarkoni: This check can move to `__init__`; that way it only runs once. (Currently it will run again each time `_extract` is called.)

PR #320, opened 2019-03-10
Adding remaining features for librosa.

jayeetaroy:
Commented  code contains changes which are in-progress and not complete.

### DIFF
--- a/pliers/extractors/audio.py
+++ b/pliers/extractors/audio.py
@@ -167,7 +167,7 @@ class LibrosaFeatureExtractor(AudioExtractor):
                 y=stim.data, sr=stim.sampling_rate, hop_length=self.hop_length,
                 **self.librosa_kwargs)
             
-        elif self._feature in[ 'tempo']:
+        elif self._feature in[ 'tempo', 'beat_track']:
             return getattr(librosa.beat, self._feature)(
                 y=stim.data, sr=stim.sampling_rate, hop_length=self.hop_length,
                 **self.librosa_kwargs)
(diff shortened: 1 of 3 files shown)


pliers/extractors/audio.py
```diff
@@ -258,6 +316,56 @@ class RMSEExtractor(LibrosaFeatureExtractor):
...
 
+class RMSExtractor(LibrosaFeatureExtractor):
+
+    ''' Extracts root mean square (RMS) from audio using the Librosa
+    library.
+
+    For details on argument specification visit:
+    https://librosa.github.io/librosa/feature.html.'''
+
+    _feature = 'rms'
```
qmac: It looks like RMS is just renamed from RMSE and RMSE will be removed in librosa `0.7.0`, https://github.com/librosa/librosa/blob/master/librosa/feature/spectral.py#L833

I'd rather just support one (RMS), and delete the code for RMSE. We will have to update `optional-dependencies.txt` to have `librosa>=0.6.3`

tyarkoni: +1

pliers/extractors/audio.py
```diff
@@ -155,27 +155,74 @@ def get_feature_names(self):
         return self._feature
 
     def _get_values(self, stim):
-        if self._feature in ['rmse', 'zero_crossing_rate']:
+        if self._feature in ['rmse', 'zero_crossing_rate', 'rms', 'spectral_flatness']:
```
tyarkoni: See @qmac's comment—we should drop `'rmse'` in favor of `'rms'`, and pin the librosa version number when this changed in our requirements.

pliers/extractors/audio.py
```diff
@@ -387,3 +495,37 @@ def __init__(self, win_length=384, **kwargs):
 
     def get_feature_names(self):
         return ['tempo_%d' % i for i in range(self.win_length)]
+
+
+    
+class TimeStretchExtractor(LibrosaFeatureExtractor):
```
tyarkoni: This and `PitchShift` will need to be `Converter` classes rather than `Extractor` classes. They should return a new `AudioStim`.

tyarkoni:
Test failures are unrelated, so merging. Thanks!

Issue #323, opened 2019-03-22
Drop Python 2 support


tyarkoni:
It's about that time. We don't need to actively do anything other than drop mention of Python 2 from the docs and remove Py2 from the travis matrix. Then we can just remove Python 2-specific features as we come along them.

qmac:
Lingering: use of `contextlib2`

tyarkoni:
Yeah, there are lots of Py2 features still having around. I closed this to reflect the fact we no longer officially support/test 2; anything that doesn't actively break on 3 can be removed at our leisure (probably as people work on related code).

PR #324, opened 2019-03-23
Minor improvements

### LINKED ISSUE #321: LibrosaFeatureExtractor can't be called directly for extractors that return multiple columns
In principle, users can invoke librosa feature extractors either via specific subclasses (e.g., `ChromaSTFTExtractor`) or via the base class (i.e., `LibrosaFeatureExtractor(feature='chroma_stft')`). In practice, the latter approach breaks for extractors that return multiple columns, because the correct `get_feature_names()` method is implemented in the subclass and not the base class (the base class's implementation just returns the name of the `feature`). The result is that when `to_df()` is called, we get an error, because the shape of the `data` don't match the shape of the `columns`.

Since we currently don't allow `LibrosaFeatureExtractor` to take more than one feature name as input (and doing so would be difficult given the variable arguments), there isn't really any benefit to calling the base class directly at the moment. Rather than invest a lot of work into this, we can just make `LibrosaFeautureExtractor` an abstract class, and force all requests to go through the subclasses.

### LINKED ISSUE #322: Abstract byte encoding of Stim data into the Stim hierarchy
Many of our API transformers need to send byte arrays of image or video data. We've written code for this enough times that it probably makes sense to abstract it into the `Stim` hierarchy itself. Probably we can just add a .`bytes` @property to relevant stims that returns a base64 encoding of the byte array (with an optional flag to turn off the encoding and return the raw data). It would also probably make sense to cache the returned value.

tyarkoni:
Includes a bunch of minor fixes/improvements: 
* Fixes style issues in recent PRs
* Makes `LibrosaFeatureExtractor` an abstract class (closes #321)
* Introduces a new `get_bytestring()` method to `ImageStim` and `VideoStim` classes that returns the byte stream encoded as a string (closes #322)
* Fixes various API-related tests that have broken over time (mostly due to changes in the results returned)
* Replaces a broken image we were using for a bunch of testing
* Replaces deprecated `scipy.misc.imread` with corresponding version from `imageio` (and fixes downstream issues in IndicoIO, which doesn't recognize ndarray subclasses)
* Starts the long-term work of dropping Python 2 support (#323) by removing `six` and various other Python 2 references as they're encountered
* Updates travis-CI config to include Python 3.6 and 3.7 in the matrix, while dropping 2.7


optional-dependencies.txt
```diff
@@ -4,8 +4,9 @@ face_recognition
 python-twitter
 gensim
 google-api-python-client
+google-compute-engine
```
qmac: Just curious, where is this used?

qmac: Nevermind, I see you just ported from the travis file.

tyarkoni: Yeah I wasn't actually sure it was used anywhere, but figured there was probably a reason it was in the travis file.

PR #326, opened 2019-03-25
First pass of the AWS Rekognition apis

tyarkoni:


pliers/extractors/api/awsReko_test.py
```diff
@@ -0,0 +1,82 @@
...
+import pandas as pd
+import boto3
+from pliers.utils import attempt_to_import, verify_dependencies
+
+aws_rekognition_client= attempt_to_import('boto3')
+
+
+class AwsRekognitionExtractor():
+    
+    def __init__(self,profile_name='default', region_name='us-east-1' ):
```
tyarkoni: `profile_name` and `region_name` should probably both be `None` by default, because both can be preconfigured in a credentials file and/or environment variables.

pliers/extractors/api/awsReko_test.py
```diff
@@ -0,0 +1,82 @@
...
+            verify_dependencies(['boto3'])
+            self.session = boto3.Session(profile_name='user2')
```
tyarkoni: Assuming the user's credential file is set up correctly, we should be able to just call `boto3.client('rekognition')`, I think. We should probably handle both cases—i.e., (a) where `profile_name` is `None`, in which case you just initialize directly, and (b) where `profile_name` is passed, in which case you initialize the session explicitly.

pliers/extractors/api/awsReko_test.py
```diff
@@ -0,0 +1,82 @@
...
+from pliers.extractors.base import Extractor, ExtractorResult
+import numpy as np
+import pandas as pd
+import boto3
+from pliers.utils import attempt_to_import, verify_dependencies
+
+aws_rekognition_client= attempt_to_import('boto3')
+
+
+class AwsRekognitionExtractor():
```
tyarkoni: Don't forget to inherit from the appropriate class—probably just the base `Extractor` class in this case.

pliers/extractors/api/awsReko_test.py
```diff
@@ -0,0 +1,82 @@
...
+                'Bytes': img2byte
+                },
+                    
+            Attributes=['ALL']
+        )
+        return response
+    
+    
+    @property
+    def img_to_byte(img):
```
tyarkoni: This will work for now, but I have an open PR I'm hoping to merge in the next couple of days that adds this functionality to the `Stim` classes directly. Once it's merged, you'll be able to do `data  = img.get_bytestring()` instead of having to include and call the custom `img_to_byte` method.

pliers/extractors/api/awsReko_test.py
```diff
@@ -0,0 +1,82 @@
...
+    #     Convert image to bytes
+        with img.get_filename() as filename:
+            with open(filename, "rb") as image:
+                f = image.read()
+                b = bytearray(f)
+                return(b)
+                
+   
+    def to_df(self, d,result=None,index=None,Key=None):
```
tyarkoni: It would be good to get in the habit of running your code against PEP8. You can user the standalone `pylint` package, but it's easier to use an IDE that supports linting plugins. Here the issue is the lack of spacing between arguments.

tyarkoni: For consistency with the `Extractor` API, I believe this needs to be `_to_df`, not `to_df`.

tyarkoni: A couple of problems with the signature of this method:
* `result` must be the first argument, and it's mandatory.
* What are `d`, `index` and `Key`? The `_to_df` method is going to be invoked from the base `Extractor` class's `to_df` method (see documentation [here](https://github.com/tyarkoni/pliers/blob/master/pliers/extractors/base.py#L110)). So you can't guarantee that anything other than `result` will be passed. (And it's unlikely users will ever pass these arguments, as they can't be seen from outside the package—i.e., only someone who knows the internals well would even know these arguments are available.) Probably best to avoid relying on them.

tyarkoni: It looks like you're trying to build a dict recursively in this method. That's fine, and is probably a good way to solve this problem, but the recursive method needs to be split out from `_to_df` and invoked explicitly the first time. Otherwise this will fail as soon as it's embedded in the `Extractor` machinery. Remember, you can't modify the `_to_df` signature; it has to be `_to_df(result)`.

adelavega:
I guess we should close this now, especially since we're going away from APIs?

tyarkoni:
Yeah, seems reasonable.

PR #330, opened 2019-03-28
[WIP/Ignore] pilmode error

adelavega:
Ross and I have been having some issues with `imageio` and `Pillow`, so I'm opening up this insignificant PR just to re-run travis tests to see its a general issue with a new version that was released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 # pliers
 
-A Python 3 package for automated feature extraction.
+A Python 3 package for automated feature extraction
 
 ## Status
 * [![Build Status](https://travis-ci.org/tyarkoni/pliers.svg?branch=master)](https://travis-ci.org/tyarkoni/pliers)


adelavega:
This is the error btw:
```
self = <imageio.plugins.pillow.JPEGFormat.Reader object at 0x7f9796f86630>
format = <Format JPEG-PIL - JPEG (ISO 10918)>
request = <imageio.core.request.Request object at 0x7f9796f865c0>
    def __init__(self, format, request):
        self.__closed = False
        self._BaseReaderWriter_last_index = -1
        self._format = format
        self._request = request
        # Open the reader/writer
>       self._open(**self.request.kwargs.copy())
E       TypeError: _open() got an unexpected keyword argument 'pilmode'
```

tyarkoni:
Not sure this should be closed. I was getting this error too—but only under fairly weird conditions (e.g., the second time tests are run). I haven't had time to look at it yet. The proximal cause for this is that scipy dropped support for `imread`, so I switched to the `imageio` version, but `imageio`'s implementation requires passing `pilmode` rather than `mode` when reading images with `pillow`.

What's not clear is why `imageio` then complains about `pilmode`. I'm guessing something in the stack is being overwritten by incompatible code at some point in the execution, leading to problems the second time through. Haven't had time to take a deep dive though.

adelavega:
I'll open an issue instead, this PR didn't do much except re-run the tests. 

Yes, I saw that `scipy` change and figured that was the source of it. I also suspect there's a versioning issue going on here. I was planning on diving in since all of neuroscout's tests are failing bc of this.

PR #332, opened 2019-03-29
Bump imageio verson to accept pilmode

### LINKED ISSUE #331: ImageIO/Pillow unexpected keyword "pilmode"
```
self = <imageio.plugins.pillow.JPEGFormat.Reader object at 0x7f9796f86630>
format = <Format JPEG-PIL - JPEG (ISO 10918)>
request = <imageio.core.request.Request object at 0x7f9796f865c0>
    def __init__(self, format, request):
        self.__closed = False
        self._BaseReaderWriter_last_index = -1
        self._format = format
        self._request = request
        # Open the reader/writer
>       self._open(**self.request.kwargs.copy())
E       TypeError: _open() got an unexpected keyword argument 'pilmode'
```

See #330

adelavega:
Closes #331

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,4 +9,3 @@ scipy>=0.13.3
 imageio>=2.3.0
 six
 tqdm
-contextlib2
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ pandas>=0.20.0
 pillow
 requests
 scipy>=0.13.3
-imageio
+imageio>=2.3.0
 six
 tqdm
 contextlib2


adelavega:
Might be better to bump even further to `2.5` (would be in line w/ `moviepy`)

tyarkoni:
I'm not sure this fixes the issue.  I was testing with `2.5` while working through the original issue, and was still getting sporadic failures...

adelavega:
Are you sure it wasn't over-ridden by another library though? My issue was that even though I had it set to `2.5` in requirements, it somehow didn't get installed due to another library's requirements. I'm testing that now.

adelavega:
(Also, regardless I think the minimum should be set to 2.5, since the code as is written requires it)

adelavega:
Okay, neuroscout tests are passing now after bumping movie py to `0.2.3.5`. It seems an older version was "incompatible" with newer imageio versions, thus not allowing it install.

tyarkoni:
LGTM except we don't need to include `contextlib2` because we're no longer officially supporting Python 2.

adelavega:
OK I removed it! I don't have write access so feel free to merge when you can.

tyarkoni:
Thanks!

PR #337, opened 2019-04-17
Add thresholds to GoogleVideoIntellgienceAPI

### LINKED ISSUE #336: Add support for frameConfidenceThreshold LabelDetectionConfig in GoogleVideoIntellgience
There are a few additional config options that could be useful, which are not currently supported:
https://cloud.google.com/video-intelligence/docs/reference/rest/v1p2beta1/videos/annotate#Feature

In particular, `videoConfidenceThreshold` (which defaults to 0.3). It could be useful to set this value lower to get a larger set of features (since the specific features cannot be specified, like in the regular image label detection)

https://github.com/tyarkoni/pliers/blob/c17cfcd45678acea381bb4df8e50ed1857437d7c/pliers/extractors/api/google.py#L325

adelavega:
Closes #336

### DIFF
--- a/pliers/extractors/api/google.py
+++ b/pliers/extractors/api/google.py
@@ -317,16 +317,37 @@ class GoogleVideoAPILabelDetectionExtractor(GoogleVideoIntelligenceAPIExtractor)
 
     ''' Extracts image labels using the Google Video Intelligence API '''
 
-    def __init__(self, mode='SHOT_MODE', stationary_camera=False, segments=None,
-                 timeout=90, request_rate=5, discovery_file=None,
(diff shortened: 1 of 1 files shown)


adelavega:
As a related idea, given that `GoogleVideoIntelligence` may be faster, and more accurate than single frames using `GoogleVisionAPI`, it may be more accurate, and faster to ask GoogleVideo for labels on evenly sampled "segments". Would it be outside of the scope of pliers to add a helper that would ask for 1hz segments based on the clip?

tyarkoni:
LGTM, thanks!

PR #339, opened 2019-04-24
Expanded/improved predefined dictionary list

tyarkoni:
Updates and expands the predefined dictionary list:

* SUBTLEXus now points to a newer version that includes extra columns
* Adds Calgary Semantic Decision Project norms (Pexman et al., 2017)
* Adds Massive Auditory Lexical Decision norms (Tucker et al., 2019)
* Adds `column_descriptions` field to some dictionaries, though this isn't used anywhere internally right now
* Fixes handling of dictionaries with duplicate index entries
* Adds `force_retrieve` flag to `PredefinedDictionaryExtractor` that forces re-download of source dictionaries


adelavega:
Can I start extract some features with this?

PR #341, opened 2019-04-30
add: tensorflow-keras-based image classifier

kaczmarj:
This classifier uses the keras `model.fit` api and predicts on multiple images in the same process. This makes it *much* faster than the current `TensorFlowInceptionV3Extractor`, which calls a separate python script on each image. The class added by this commit only has to do the TensorFlow setup things once, so only the first predict is impacted. All subsequent predictions are fast.

The model comes from https://github.com/keras-team/keras-applications and the weights are pre-trained on imagenet.

Here is code to test (in ipython):

```python
import os
from pliers.extractors import TensorFlowInceptionV3Extractor
from pliers.extractors import TensorFlowKerasInceptionV3Extractor
from pliers.filters import FrameSamplingFilter
from pliers.stimuli import VideoStim

video = os.path.join(get_test_data_path(), 'video', 'obama_speech.mp4')

sampler = FrameSamplingFilter(hertz=1)
frames = sampler.transform(video)
ext = TensorFlowInceptionV3Extractor()
%time results = ext.transform(frames)

sampler = FrameSamplingFilter(hertz=1)
frames = sampler.transform(video)
ext_new = TensorFlowKerasInceptionV3Extractor()
%time results_new = ext_new.transform(frames)
```

On my computer (with intel i5 processor), this is the timing output for the original class:

```
Stim: 9it [01:17,  8.76s/it]
CPU times: user 1.75 s, sys: 49.9 ms, total: 1.8 s
Wall time: 1min 17s
```

and this is the output for the new class:

```
Stim: 9it [00:02,  5.25it/s]
CPU times: user 4.26 s, sys: 514 ms, total: 4.77 s
Wall time: 2.2 s
```


qmac:
Nice! I'm tempted to say we should retire the `TensorFlowInceptionV3Extractor`

tyarkoni:
Thanks @kaczmarj, this is awesome! I'll review this shortly.

I agree with @qmac that this effectively obsoletes `TensorFlowInceptionV3Extractor`; do you mind adding a commit removing it from the codebase?

Also, on a tangent, I should note that I have a working local branch for generic feature extraction using any named language model in TF Hub. There's no overlap in effort here (and I don't think I've implemented support for images yet); just noting that in case you have any ambitions to expand TF support (which is definitely high on my agenda!).

pliers/extractors/models.py
```diff
@@ -84,3 +84,65 @@ def _extract(self, stim):
             values.append(float(extraction[1]))
 
         return ExtractorResult([values], stim, self, features=features)
+
+
+def _resize_image(image, shape):
```
tyarkoni: This can probably wait, but I imagine that changes in aspect ratio will tend to throw the model off, so ideally the resizing here should maintain the aspect ratio of the input and pad with empty borders as needed. I don't remember how or if this was handled in the previous extractor (I certainly didn't do it myself, but perhaps it's done internally).

pliers/extractors/models.py
```diff
@@ -84,3 +84,65 @@ def _extract(self, stim):
...
+    def __init__(self, weights=None, num_predictions=5):
+        super(TensorFlowKerasInceptionV3Extractor, self).__init__()
+        if weights is None:
+            weights = 'imagenet'
+        self.weights = weights
+        self.num_predictions = num_predictions
+
+        import tensorflow as tf
+        # Instantiating the model also downloads the weights to a cache.
+        self.model = tf.keras.applications.inception_v3.InceptionV3(
```
tyarkoni: I haven't looked into this in any detail, but if the models here all have the same API, I wonder if we can just replace this with a generic `TensorFlowKerasApplicationExtractor` that takes arbitrary init arguments, and then wrap that with very thin subclasses for the individual extractors? I'm not suggesting you should do this yourself, @kaczmarj, but just wondering whether you think this is a viable approach (in which case we can open an issue for a refactor along these lines).

pliers/extractors/models.py
```diff
@@ -84,3 +84,65 @@ def _extract(self, stim):
...
+    VERSION = '1.0'
+
+    def __init__(self, weights=None, num_predictions=5):
+        super(TensorFlowKerasInceptionV3Extractor, self).__init__()
+        if weights is None:
+            weights = 'imagenet'
+        self.weights = weights
+        self.num_predictions = num_predictions
+
+        import tensorflow as tf
```
tyarkoni: We have a convention for importing optional dependencies that allows us to detect missing libraries; see the `attempt_to_import` and `verify_dependencies` utilities used in other extractors.

pliers/extractors/models.py
```diff
@@ -84,3 +84,65 @@ def _extract(self, stim):
...
+def _resize_image(image, shape):
+    import numpy as np
+    from PIL import Image
+    return np.array(
+        Image.fromarray(image).resize(shape, resample=Image.BILINEAR))
+
+
+class TensorFlowKerasInceptionV3Extractor(ImageExtractor):
+
+    ''' Labels objects in images using a pretrained Inception V3 architecture
```
tyarkoni: Maybe add a note that images must be 299x299x3, and will be resampled (with distortion) if a different input is passed. That way at least people are aware that they might get wonky results unless they do some editing.

pliers/extractors/models.py
tyarkoni: On further thought, in principle, I think this is exactly the kind of problem that we have the `Filter` hierarchy for... it makes sense to me to implement an `ImageResizingFilter` that basically wraps this function and optionally adds padding etc. Again, not so much a suggestion for present action as a note to self to do this in future (and may be worth opening an issue for).

pliers/extractors/models.py
```diff
@@ -84,3 +84,65 @@ def _extract(self, stim):
...
+
+        required_shape = (299, 299, 3)
+        x = stim.data
+        if x.ndim != 3:
+            raise ValueError("Stim must have rank 3.")
+        if x.shape != required_shape:
+            x = _resize_image(x, required_shape[:-1])
+        x = tf.keras.applications.inception_v3.preprocess_input(x)
+        x = x[None]  # Add batch dimension.
+        preds = self.model.predict(x, batch_size=1)
```
tyarkoni: Not sure you want to go to the trouble, but in principle, there's a [`BatchTransformerMixin`](https://github.com/tyarkoni/pliers/blob/b9b843c3e383cea2f5d774d471b5499c4038f7d4/pliers/transformers/base.py#L217) that allows you to handle multiple `Stim` objects within `_extract` rather than naively looping. Not sure if it would make a meaningful difference in efficiency in this case.

pliers/extractors/models.py
tyarkoni: See the [`LibrosaFeatureExtractor`](https://github.com/tyarkoni/pliers/blob/master/pliers/extractors/audio.py#L140) hierarchy for an example of what this might look like.

tyarkoni:
Oh, and do you mind adding a test for this? We try to have 100% coverage for new extractors before merging (by number of extractors, I mean, not necessarily LOC).

pliers/extractors/models.py
kaczmarj: I haven't tested this, but I don't think changing the aspect ratio should affect the prediction too much. The model is definitely trained with augmentation, which includes distorting the aspect ratios of the images. In the tensorflow/keras [tutorial](https://www.tensorflow.org/tutorials/images/hub_with_keras#run_it_on_a_single_image) on image classification, they use the Pillow resize method, which doesn't maintain aspect ratio. That being said, I definitely can add an `ImageResizingFilter` that can resize while optionally maintaining aspect ratio. I have the code written, but I'll submit a different PR for it.

pliers/extractors/models.py
kaczmarj: The models do share the same API (go keras!). We probably wouldn't even need subclasses. The differences across models will be their location in the applications package as well as the required input shape. A simple mapping of these things would probably suffice. I can do that too (in a separate PR).

pliers/extractors/models.py
kaczmarj: We could do this, but there won't be a meaningful improvement in performance. In my own work, I've found that predicting on single samples in a loop is just as fast as predicting on batches. I don't know why that would be though...

kaczmarj:
huh, the test failed with the error

```python
assert 0.9737075 == 0.9737075
```

must be a floating point thing. i'll test with `np.allclose`.

pliers/extractors/models.py
qmac: That seems odd, just curious what size batches were you testing?

kaczmarj: The batches were small. This was for a 3D U-Net that worked on inputs of (128, 128, 128, 1). I found that predicting on single samples was as fast as predicting on batches of 4 (I couldn't go any higher due to memory constraints).

tyarkoni:
LGTM! Okay to merge, @kaczmarj?

kaczmarj:
Good to merge on my end! Thanks

tyarkoni:
Awesome, thanks again!

pliers/extractors/models.py
kaczmarj: #342 adds the `ImageResizingFilter`

pliers/tests/extractors/test_image_extractors.py
```diff
@@ -54,16 +54,19 @@ def test_saliency_extractor():
...
     results = ext.transform(imgs)
     df = merge_results(results, format='wide', extractor_names='multi')
-    assert len(df) == 2
-    assert ('TensorFlowInceptionV3Extractor', 'Granny Smith') in df.columns
-    assert 0.22610 in df[
-        ('TensorFlowInceptionV3Extractor', 'Windsor tie')].values
+    assert df.shape == (2, 19)
+    true = 0.9737075
+    pred = df['TensorFlowKerasInceptionV3Extractor'].loc[0, 'Granny_Smith']
+    np.isclose(true, pred, 1e-05)
```
qmac: Does this need an `assert` to work?

kaczmarj: yes it does...

tyarkoni: Oh, hah, good catch. :)

PR #342, opened 2019-05-02
Add: ImageResizeFilter

kaczmarj:
This PR proposes adding an image resizing filter, which can optionally maintain aspect ratio. This is following up https://github.com/tyarkoni/pliers/pull/341#discussion_r279959595.

Here is some test code:

```python
import os
from pliers.filters import ImageResizingFilter
from pliers.stimuli import ImageStim

image_path = os.path.join(get_test_data_path(), 'image', 'apple.jpg')
from PIL import Image

# Shape of the image is (288, 420, 3)
image = ImageStim(filename=image_path)

resizer = ImageResizingFilter((299, 299), maintain_aspect_ratio=False)
new_stim = resizer.transform(image)
print(new_stim.data.shape)
Image.fromarray(new_stim.data)

resizer = ImageResizingFilter((299, 299), maintain_aspect_ratio=True)
new_stim = resizer.transform(image)
print(new_stim.data.shape)
Image.fromarray(new_stim.data)
```

Output of first resizer:
image

Output of second resizer:

image

### DIFF
--- a/pliers/filters/image.py
+++ b/pliers/filters/image.py
@@ -53,20 +53,31 @@ class ImageResizingFilter(ImageFilter):
             maintaining aspect ratio, and pad the rest with zero values.
             Otherwise, potentially distort the image during resizing to fit the
             new size.
-        resample (int): resampling method. 0 for nearest neighbor, 1 for
-            Lanczos, 2 for bilinear, 3 for bicubic (default), 4 for box, and 5 for
-            hamming. See https://pillow.readthedocs.io/en/5.1.x/handbook/concepts.html#concept-filters
+        resample str: resampling method. One of 'nearest', 'bilinear',
+            'bicubic', 'lanczos', 'box', and 'hamming'. See
+            https://pillow.readthedocs.io/en/5.1.x/handbook/concepts.html#concept-filters
             for more information.
     '''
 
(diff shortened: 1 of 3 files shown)


pliers/filters/image.py
```diff
@@ -43,6 +43,58 @@ def _filter(self, stim):
...
+
+    ''' Resizes an image, while optionally maintaining aspect ratio.
+
+    Args:
+        size (tuple of two ints): new size of the image.
+        maintain_aspect_ratio (boolean): if true, resize the image while
+            maintaining aspect ratio, and pad the rest with zero values.
+            Otherwise, potentially distort the image during resizing to fit the
+            new size.
+        resample (int): resampling method. 0 for nearest neighbor, 1 for
```
tyarkoni: I find `int` codes for things like this suboptimal from a UX standpoint, because it forces the user to read the docstring every time they want to change the setting. Maybe replace with `['nn', 'lanczos', 'bilinear', 'bicubic', 'box', 'hamming']` options? I get that the wrapped pillow method accepts ints as the same argument, but they also have mapped constants like `PIL.Image.BICUBIC` which are easier to remember. It would be fine to accept both (and say so in the docstring)—i.e., either one of the above strings, or any constant/integer pillow's resize can handle.

PR #343, opened 2019-05-02
ENH: dockerfile

kaczmarj:
I tried building a pliers docker image with the current Dockerfile, but it raised an error during the build. So I decided to refactor the image. This introduces some major changes into the image, but I think it streamlines it nicely. The docker image in this PR bootstraps `python:3.6-slim` and installs all dependencies from pip.

This image also adds a non-root user, so you can use jupyter without having to pass in `--allow-root`, and things are a bit safer in general.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -13,7 +13,9 @@ RUN apt-get update -qq \
     && pip install --no-cache-dir \
         --requirement requirements.txt \
         --requirement optional-dependencies.txt \
+    && python -m spacy download en_core_web_sm \
     && pip install --no-cache-dir --editable . \
+    && rm -rf ~/.cache/pip \
     && apt-get autoremove --purge -yq $tmp_pkgs \
     && apt-get clean \
     && rm -rf /var/lib/apt/lists/* \
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,33 +1,22 @@
-FROM continuumio/miniconda3:4.5.11
-
+FROM python:3.6-slim
 ARG DEBIAN_FRONTEND="noninteractive"
-
-RUN apt-get -qq update \
-    && apt-get install -yq --no-install-recommends \
-      git \
-      graphviz \
-      libgraphviz-dev \
-      pkg-config \
-      tesseract-ocr \
-    && apt-get clean \
-    && rm -rf /var/lib/apt/lists/*
-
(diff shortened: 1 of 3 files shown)


tyarkoni:
Awesome, thanks! LGTM, but I'm fairly Docker-inexperienced. @poldrack, since you contributed the previous version, do you want to take a quick look?

docker/Dockerfile
```diff
@@ -1,42 +1,26 @@
...
+        graphviz \
+        libmagic1 \
+        tesseract-ocr \
+        $tmp_pkgs \
+    && pip install --no-cache-dir \
+        --requirement requirements.txt \
+        --requirement optional-dependencies.txt \
+        ipython \
+        notebook \
+    && python -m spacy download en_core_web_sm \
```
adelavega: This may also be needed:
`python -m pliers.support.download`

Also, I think it might be better to split this huge `RUN` command into several smaller `RUN` commands, rather than one big statement. Docker build caching might work better (unless you have a reason for doing it this way?

kaczmarj: I was debating whether to add `python -m pliers.support.download`. That downloads things into the current `$HOME`, so if someone used the container as a different user (e.g., in singularity), those assets wouldn't be available. But I can add that download, and users can re-download if they do end up using the container as a different user.

I suppose parts of the `RUN` instruction can be split, but I wrote it this way because I download and later delete temporary dependencies (e.g.,`gcc`). If the download and deletion were in separate `RUN` instructions, the size of the resulting image would be larger.

I also ran `pytest` in the container, and the tests that don't require payment passed.

adelavega: Cool, that's a good reason to keep it as is then!

Just add the support download (I think its best if the image works for most users out of the box)

adelavega:
Hey @kaczmarj what command do you use to build the `Dockerfile`?

docker/Dockerfile
kaczmarj: Sounds good. And I think I did need to run `python -m pliers.support.download` for the tests to pass.

kaczmarj:
@adelavega - to build the docker image, go to the root pliers directory and run

```
docker build --tag pliers -f docker/Dockerfile .
```

adelavega:
Thanks! Please update `docker/README` please. I think that's outdated then.

tyarkoni:
Thanks!

Issue #220, opened 2017-12-29
Explicit __init__ methods for Transformer classes


tyarkoni:
A minor issue with the current `Transformer` hierarchy is that most of the user-facing classes don't have to define explicit `__init__` methods. This isn't a problem functionally, but it makes it hard for users to figure out what the initialization arguments are--and this can be quite important (e.g., in the case of arguments like `api_key`). Even though it's redundant, we should probably insist on every public-facing `Transformer` class explicitly implementing `__init__` with all the main arguments visible, even if all the method does is turn right around and pass everything up to the superclass.

qmac:
I agree with this issue as well, it just will be tedious to do everything. This also leaves us open to interpreting what "main arguments" are, but I suspect we'll just have to use our judgment per transformer. Do the `LibrosaFeatureExtractor`s look good for this pattern?

tyarkoni:
Yeah I think we probably need to do the librosa Extractors this way too. But there's no rush on this--it's not a high priority, and we can chip away at it as time allows. It was mainly salient to me because I'm close to pushing a big PR that adds Sphinx docs, and the autodocs don't help very much when the args are basically just `**kwargs`.

PR #344, opened 2019-05-04
ENH: replace tensorflow extractor with more generic version.

kaczmarj:
This class accepts a model architecture string, and will use that pre-trained model from `tensorflow.keras.applications`. These changes also check whether `tensorflow.keras` is available. For image resizing, `pliers.filters.ImageResizingFilter` is used. The architecture 'inceptionv3' is the default.

Here is code for testing:

```python
import os
from pliers.extractors import TensorFlowKerasApplicationExtractor
from pliers.stimuli import ImageStim

path = os.path.join(get_test_data_path(), 'image', 'apple.jpg')
image = ImageStim(path)
ext = TensorFlowKerasApplicationExtractor(architecture='inceptionv3')
result = ext.transform(image)
```

### DIFF
--- a/pliers/extractors/models.py
+++ b/pliers/extractors/models.py
@@ -1,67 +1,106 @@
 ''' Extractor classes based on pre-trained models. '''
 
 import numpy as np
-from PIL import Image
 from pliers.extractors.image import ImageExtractor
 from pliers.extractors.base import ExtractorResult
+from pliers.filters.image import ImageResizingFilter
 from pliers.utils import attempt_to_import, verify_dependencies
 
 
 tf = attempt_to_import('tensorflow')
+attempt_to_import('tensorflow.keras')
 
 
-def _resize_image(image, shape):
-    return np.array(
-        Image.fromarray(image).resize(shape, resample=Image.BICUBIC))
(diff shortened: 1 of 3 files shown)


pliers/extractors/models.py
```diff
@@ -1,67 +1,106 @@
...
+    def __init__(self,
+                 architecture='inceptionv3',
+                 weights=None,
+                 num_predictions=5):
         verify_dependencies(['tensorflow'])
-        super(TensorFlowKerasInceptionV3Extractor, self).__init__()
+        verify_dependencies(['tensorflow.keras'])
+        super(TensorFlowKerasApplicationExtractor, self).__init__()
+
+        # Model name: (model module, model function, required shape).
```
qmac: Consider using a NamedTuple or some structure to avoid needing hard-coded indexing later. Not a big deal though.

tyarkoni:
Re: subclasses, I agree that would be nice in principle, but to my mind the main benefit would be maintaining separate docstrings/init args (see #220). In this case it looks like all the models take the same arguments, so I think there's less of a reason to go this route.

PR #346, opened 2019-05-09
ENH: precreate non standard /data /backup directories in Dockerfile

yarikoptic:
This is to facilitate reuse of the docker image when fetched by singularity
on a system with an elderly kernel built without support of overlays and thus
requiring those directories (present on our HPC) to be present within image

No hurt feelings if rejected -- just wanted to check for an easy way out for myself ;)


tyarkoni:
Fine with me—any objections, @kaczmarj?

kaczmarj:
Looks great to me

PR #348, opened 2019-05-09
ENH: Dockerfile - chmod to guarantee availability, and test global installation via import pliers

yarikoptic:
the first call `python -m pliers.support.download` is done within source code tree, so mileage could vary from calling from outside

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,6 +2,7 @@ FROM python:3.6-slim
 ARG DEBIAN_FRONTEND="noninteractive"
 WORKDIR /opt/pliers
 COPY . .
+RUN chmod a+rX -R .
 RUN apt-get update -qq \
     && tmp_pkgs="cmake gcc g++ libc6-dev libgraphviz-dev libmagic-dev make" \
     && apt-get install -yq --no-install-recommends \


docker/Dockerfile
```diff
@@ -28,3 +29,5 @@ RUN mkdir -p /data /backup
 USER pliers
 RUN python -m pliers.support.download
 WORKDIR /work
+# Just a test to guarantee correct installation
+RUN python -m pliers
```
kaczmarj: @yarikoptic - Would `RUN python -m pliers.support.download` not already test that the installation is correct?

yarikoptic: It is ran within source code base. In principle local module shouldn't be picked up, but I am paranoid, so thought it wouldn't hurt to test from a work dir

yarikoptic: But thinking about it, -m pliers shouldn't be correct if there is no `__main__` there... Let me check later...

yarikoptic: damn me -- I have managed to miss (was running build on another box) that it did fail as then I feared:
```
Step 11/11 : RUN python -m pliers
 ---> Running in c5405074bfef
/usr/local/bin/python: No module named pliers.__main__; 'pliers' is a package and cannot be directly executed
The command '/bin/sh -c python -m pliers' returned a non-zero code: 1
```
I will just remove that "Test" change for now.  Ideally it should just run all the tests upon build to verify correct operation in the environment it has created.  but let's keep it a separate issue ;)

yarikoptic:
ready for the merge now ;) thanks!

PR #351, opened 2019-05-11
Add description to affect dataset

adelavega:
Let me know if this is useful, and if so I can add them for the other datasets. 
Main issue I see is its a bit verbose. 

Either way, I have to write something for NS.


adelavega:
Actually, for NS I'll probably use something more verbose since I'm only going to extract the Mean ratings.

PR #355, opened 2019-08-07
[WIP] safeguard against keyerror in clarifai face detection

adswa:
Toying around with the face detection models of clarifai, I noticed that while the extraction works, the transformation into a dataframe with ``_to_df()`` fails due to KeyErrors.
This is a very basic safe guard that checks whether the model used for the clarifai client is a Face Detection model (they identify it by this character string) and if so, returns the face region boundaries as a dataframe if a face is present and nothing if no face is present.

I had plenty of problems testing all if this locally, because I exceeded my free request limit, so I'm PRing a work in progress for further discussion about alternatives, or any other improvements.

### DIFF
--- a/pliers/extractors/api/clarifai.py
+++ b/pliers/extractors/api/clarifai.py
@@ -98,7 +98,7 @@ class ClarifaiAPIExtractor(APITransformer):
     def _parse_annotations(self, annotation):
         data_dict = {}
         # check whether the model is the face detection model
-        if self.model_name == 'a403429f2ddf4b49b307e318f00e528b':
+        if self.model_name == 'face':
             # if a face was detected, get at least the boundaries
             if annotation['data']:
                 for k, v in annotation['data']['regions'][0]['region_info']['bounding_box'].items():
--- a/pliers/extractors/api/clarifai.py
+++ b/pliers/extractors/api/clarifai.py
@@ -97,8 +97,16 @@ class ClarifaiAPIExtractor(APITransformer):
 
     def _parse_annotations(self, annotation):
         data_dict = {}
-        for tag in annotation['data']['concepts']:
-            data_dict[tag['name']] = tag['value']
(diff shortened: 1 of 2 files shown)


pliers/extractors/api/clarifai.py
```diff
@@ -97,8 +97,16 @@ def _query_api(self, objects):
 
     def _parse_annotations(self, annotation):
         data_dict = {}
-        for tag in annotation['data']['concepts']:
-            data_dict[tag['name']] = tag['value']
+        # check whether the model is the face detection model
+        if self.model_name == 'a403429f2ddf4b49b307e318f00e528b':
```
tyarkoni: Do we know that this is a static label for the face model? I'm worried that this is just a hash they use to encode the specific version of the model, and that it might change frequently. It might be better to rely on the model name stored in `self`.

adswa:
Ha, thx @tyarkoni for that thought! I did finally find the model name they are using. Other than stated on their [website](https://www.clarifai.com/models/face-detection-image-recognition-model-a403429f2ddf4b49b307e318f00e528b-detection), the model name is simply 'face', and I missed that there is a second `model_name` attribute within `self.model`.

tyarkoni:
Great! Do you mind adding a test for this in `test_clarifai_extractors.py`? You can adapt from one of the existing tests and just check to make sure some of the face-specific features are returned successfully in the `ExtractorResult`. Thanks!

adswa:
Good point - I added some basic tests, let me know whether they are sufficient.

luciajajcay:
I'm also trying to use the face model, have used both ext = ClarifaiAPIImageExtractor(model='a403429f2ddf4b49b307e318f00e528b') and ext = ClarifaiAPIImageExtractor(model='face') but keep getting a key error (KeyError: 'concepts') when trying to transform the results to a dataframe. Do you guys know how to solve this, by any chance?

PR #357, opened 2019-08-08
[WIP] Adjust clarifai face detection model for multi-face stimuli

adswa:
My previous PR (#355) adjusted the clarifai face detection model to work only with single face images. In the case of more than one image, only the last image would be returned. This PR returns all detected faces in an image or video in a multi-row data frame.
I've also added basic tests that check whether more than a single face is returned in the case of multiple faces in the stimuli.

Copying the google API face extraction, I've added a parameter to return only the first face, if desired (`handle_annotations'), but I'm not using that anywhere so far (mainly because I can't find it used for Google anywhere).


tyarkoni:
Sorry about the delay getting to this! Thanks again!

PR #358, opened 2019-08-08
Added optional dependences as 'extras' for pip install

snastase:
This adds and `extras_require` field to the `setup.py`, which should all users to optionally install all optional dependencies using `pip install pliers[all]`. This duplicates the functionality of `pip install -r optional-dependencies.txt`, but without installing from GitHub. The installation runs, but I haven't fully verified that everything works perfectly because I'm currently running into a `dlib` installation error for other reasons. I also made a small update to the installation documentation to reflect this change.


tyarkoni:
Thanks!

PR #360, opened 2019-08-19
Patch operations request for Google Video Intelligence

### LINKED ISSUE #359: Google Vision Api Change?
Not sure why but I am getting the following error with any of the Google Transformers. 
My guess is that the google api or library changed, and that since the optional dependency version is not pinned, we are out of sync:

image

adelavega:
Closes #359 

Not sure why this new syntax is required (the API should be v1, and according to documentation, it should still operate the same)

### DIFF
--- a/pliers/extractors/api/google.py
+++ b/pliers/extractors/api/google.py
@@ -205,7 +205,11 @@ class GoogleVideoIntelligenceAPIExtractor(GoogleAPITransformer, VideoExtractor):
         return request_obj.execute(num_retries=self.num_retries)
 
     def _query_operations(self, name):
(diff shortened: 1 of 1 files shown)


adelavega:
This patch fixed my issue. Tests failing for unknown reasons.

tyarkoni:
Yeah there's an intermittent issue causing segfaults unpredictably on travis. Doesn't happen for me locally and I don't have time to get to the bottom of it (doubt it will be straightforward). I'll merge this as-is seeing as it fixes the issue and is a high priority.

tyarkoni:
Thanks!

PR #361, opened 2019-09-22
Text embeddings 092019

debanjanghosh:
This draft pull request is regarding the new text embeddings that Pliers can support. Note, a thorough description of the supported features and methods are described in the attached file (TextEncoding_pliers.pdf). Pliers now can support popular text encoding such as word embeddings ("glove", "word2vec", "fasttext") and new context-embedding such as "ELMo" as well as the transformer based architecture such as "BERT".

The new features are implemented in text_encoding.py script in /pliers/extractors/ sub-directory. The associated files (e.g., code for bert-modeling or bert-based tokenization) are also in /pliers/extractors/ sub-directory. The list of the new files are below.

1. SIF_embedding.py
2. bert_extract_features.py
3. bert_modeling.py
4. bert_tokenization.py
5. sif_data_io.py
6. sif_params.py
7. sif_tree.py
8. skipthoughts.py
9. text_encoding.py

All the required embeddings and pre-trained models  are hosted on AWS. A script ("download_embedding.py") is provided in the /pliers/support/ sub-directory that will help users to download the particular required embeddding/model. Description of the function calls are described (with examples) in the attached draft. 

Pliers now also support "prefix embedding", i.e., for a sentence that contains five words it is possible to extract embeddings of the continuous n-grams (i.e., first word, first two words, first three words...). 

For testing, the test file is "test_text_embeddings.py" (in /tests/extractors/ sub-directory). An example file is also provided - "text_encoding_pliers.py" in the same folder for examples (this can be removed later).


debanjanghosh:
Also, here is the draft - TextEncoding_Pliers.pdf that describe in details all the new functionalities, their use 
[TextEncoding_Pliers.pdf](https://github.com/tyarkoni/pliers/files/3640061/TextEncoding_Pliers.pdf)
with examples.

pliers/tests/extractors/brain_decoding_text_encoding_interface.py
```diff
@@ -0,0 +1,277 @@
+import warnings
```
qmac: Can you elaborate on the purpose of this file? Does it add testing functionality beyond what is covered in your other unit test file, `test_text_embeddings.py`?

pliers/support/download_embedding.py
```diff
@@ -0,0 +1,165 @@
...
+import requests
+import sys
+import tarfile
+import logging
+
+import gensim
+from gensim.scripts.glove2word2vec import glove2word2vec,get_glove_info
+import argparse
+
+_aws_bucket_path = 'https://s3.amazonaws.com/mlt-word-embeddings/'
```
qmac: Is this bucket completely public? Does it require set of AWS credentials?

tyarkoni:
Awesome, thanks @debanjanghosh! And thanks for the review, @qmac! I'll take a look at this as well shortly.

pliers/support/download_embedding.py
tsuipen: Hi qmac, the bucket belongs to the machine learning team at NIMH, and is public. We use it to share embeddings and models with other researchers, since the files are large. Debanjan and I thought it would make sense to develop the code with a single source for all the embeddings and models, and since we already had them all in one place... We will be happy to contemplate other possibilities in the longer term.

tsuipen:
The problem there appears to be that some dependencies need to be installed, and that information is only on the companion pdf instructions. The relevant bit is

"Install these packages

gensim=3.4.0, tensorflow, theano, sklearn, nltk, allennlp


using whatever installer you prefer. If using conda, try installing with “conda install <package name>”, using “pip install <package name>” only if that fails. If the installation of allennlp fails in conda with an error about jsonnet, "conda install -c conda-forge jsonnet" prior to trying it should address it. Once installed, start an interactive python and

[1] import nltk
[2] nltk.download('stopwords') "

pliers/tests/extractors/brain_decoding_text_encoding_interface.py
debanjanghosh: @qmac, sorry this file was a part of a project run by MIT+NIMH and it should not be here. I am making a note and will remove it. Thanks.

pliers/support/download_embedding.py
qmac: Sounds good to me, thanks!

pliers/extractors/text_encoding.py
```diff
@@ -0,0 +1,812 @@
...
+
+keyedvectors = attempt_to_import('gensim.models.keyedvectors', 'keyedvectors',
+                                 ['KeyedVectors'])
+doc2vecVectors = attempt_to_import('gensim.models.doc2vec','doc2vecVectors',
+                                   ['Doc2Vec.load'])
+
+logging.getLogger('smart_open').setLevel(logging.ERROR)
+logger = logging.getLogger("text_encoding_logger") 
+
+class DirectTextExtractorInterface():
```
adelavega: The pattern here of making a base Interface class which is configurable but instantiates distinct classes for each method does not follow the current style.

What we have more commonly done is implement a base class which may not anything, or at least much, on its own, but sets the common pattern for more specific subclasses. 

A simple example is `FaceRecognitionFeatureExtractor`. This class is not imported in `__init__` because its not a complete extractor. But it is then subclassed by three extractors. When necessary, the common functions can be re-implemented by these extractors. More complex examples of this is `GoogleVisionAPIExtractor` set of extractors.

pliers/extractors/text_encoding.py
```diff
@@ -0,0 +1,812 @@
...
+
+
+        elif method.lower() == 'skipthought':
+            
+            self.semvector_object = SkipThoughtExtractor()
+            self.semvector_object._loadModel() 
+            
+        elif method.lower() == 'sif':
+            
+            self.semvector_object = SmoothInverseFrequencyExtractor(embedding=embedding,\
```
adelavega: See my above comment, but I think it makes more sense to have these other specific extractors inherit from this base class, rather than be instantiated and set as instance variables by this class. 

Ideally, this pattern maximizes re-use of parallel code, and highlights the differences between the extractors. 

Additionally, this means that each specific extractor type is available externally (i.e. in `__init__`), which is useful when making pliers extraction graphs, etc..

pliers/extractors/text_encoding.py
```diff
@@ -0,0 +1,812 @@
...
+
+
+    def _loadModel(self):
+        self.wvModel = keyedvectors.KeyedVectors.load_word2vec_format(self.embedding_file,binary=self.binary,encoding='latin1')
+        logger.info(self.embedding + ': ' + 'model loaded')
+
+    def _getModel(self):
+        return self.wvModel 
+
+    def getVector(self,stim,cbow=False):
```
adelavega: It looks like for the most part you've already set each class to have an overlapping set of common methods. This will probably make it easier to refactor as I suggest above.

pliers/extractors/text_encoding.py
```diff
@@ -0,0 +1,812 @@
...
+    _max_seq_length = 128
+    _bert_config_file = 'bert_config.json'
+    _do_lower_case = True
+    _num_tpu_cores = 8
+    _master = None
+    _use_tpu = False
+    _init_checkpoint = 'bert_model.ckpt'
+    _use_one_hot_embeddings = False
+    _method = 'bert'
+    _model = 'bert_uncased_L-12_H-768_A-12'
```
adelavega: The more that these config variables overlap in meaning between models the better.  Usually we use these to indicate variables that differentiate the behavior between similar extractors, rather than to specify config variables that only apply to a single extractor.

pliers/extractors/text_encoding.py
```diff
@@ -0,0 +1,812 @@
...
+                               features=features)
+
+ 
+class SkipThoughtExtractor(DirectSentenceExtractor):
+    
+    '''To learn about skipthought extractor please
+        refer Skip-Thought Vectors paper by Kiros et al.'''
+    
+    '''Note: skipthought is based on Theano 1.0.4 and
+        Numpy <1.16.3 (we are using numpy-1.16.2)'''
```
adelavega: So it doesn't run right now? Try bumping up numpy and see if anything breaks.

pliers/support/__init__.py
```diff
@@ -4,8 +4,10 @@
 
 from .decorators import requires_nltk_corpus
 from .download import download_nltk_data
+from .download_embedding import download_embedding_data
```
adelavega: I think this function should go in `.download`. No need for a new module.

pliers/support/download_embedding.py
adelavega: Yeah I worry a little bit about guaranteeing that these locations will always be available, but I suppose that's the case for any pre-trained model that needs to be downloaded.

pliers/support/download_embedding.py
```diff
@@ -0,0 +1,165 @@
...
+    try:
+        size = os.stat(embedding_local_file).st_size
+        logger.info('\tSuccessfully downloaded:' + '\t' + embedding_model_file)
+        tarfile.open(embedding_local_file, 'r:gz').extractall(embedding_model_full_path)
+        
+    except IOError as e:
+        logger.info(e)
+
+
+def parseArguments():
```
adelavega: I don't think a CLI should be here

pliers/extractors/SIF_embedding.py
```diff
@@ -0,0 +1,61 @@
+import numpy as np
```
adelavega: Who's the author of this code?

pliers/extractors/bert_extract_features.py
```diff
@@ -0,0 +1,544 @@
+# coding=utf-8
```
adelavega: So is all this code authored by Google AI? Is this maintained elsewhere?

pliers/extractors/bert_extract_features.py
```diff
@@ -0,0 +1,544 @@
...
+from pliers.extractors import bert_tokenization
+import tensorflow as tf
+
+import numpy as np
+
+flags = tf.flags
+
+FLAGS = flags.FLAGS
+
+flags.DEFINE_string("input_file", None, "")
```
adelavega: I am concerned about implementing extractor specific / case specific CLIs within pliers. If we do, they should at least not be in extractors, but I have a feeling that these don't belong in pliers at all. This sort of code can certainly wrap pliers and live elsewhere, or maybe we could generalize this code and not make it specific to text embeddings (and work for all of pliers).

adelavega:
I'm going to close this in favor of #365 
Thanks!

Issue #340, opened 2019-04-24
Add repetition counting extractor for ComplexTextStims


tyarkoni:
To be able to easily model the effects of word repetition in NeuroScout, we can add a `RepetitionCountExtractor` class that takes a `ComplexTextStim` as input, and outputs the number of times each word has been previously seen before (with optional log transformation of values).

adelavega:
Closed by #364

PR #364, opened 2019-12-20
adding word counter

rbroc:
Implements extractor that counts number of times a word occurs. 
Includes case_sensitive flag and option to return count in log scale.
Solves #340


adelavega:
@rbroc we'll need to add formal tests as well, something we haven't discussed much yet.

pliers/extractors/text.py
```diff
@@ -402,3 +402,33 @@ def _extract(self, stim):
 
         return ExtractorResult(features_list, stim, self,
                                features=self.features, orders=order_list)
+
+class WordCounterExtractor(TextExtractor):
+
+    ''' Extracts number of times word has occurred within text
```
tyarkoni: Maybe clarify that it's the number of times each unique word occurs–current wording is easy to interpret as matching a specified word.

pliers/extractors/text.py
```diff
@@ -402,3 +402,33 @@ def _extract(self, stim):
 
         return ExtractorResult(features_list, stim, self,
                                features=self.features, orders=order_list)
+
```
tyarkoni: Follow PEP8—leave two lines between top-level function/class definitions.

pliers/extractors/text.py
```diff
@@ -402,3 +402,33 @@ def _extract(self, stim):
...
+class WordCounterExtractor(TextExtractor):
+
```
tyarkoni: We have a convention of tracking any initialization parameters we want to log (basically, anything that helps the user understand the instance's expected behavior) using a `_log_attributes` class variable. So you could add the line `_log_attributes = ('lemmatize',)` here. See other classes in the module for examples.

pliers/extractors/text.py
```diff
@@ -402,3 +402,33 @@ def _extract(self, stim):
...
+        tokens = nltk.word_tokenize(stim.text)
+        
+        if self.lemmatize:
+            lemmatizer=nltk.WordNetLemmatizer()
+            pos_map = dict(zip(['ADJ', 'ADJ_SAT', 'ADV', 'NOUN', 'VERB'], ['a', 's', 'r', 'n', 'v'])) # map tags to wordnet tagset
+            tokens = dict(nltk.pos_tag(tokens, tagset='universal'))
+            tokens = {k: pos_map[v] if v in pos_map else 'n' for k, v in tokens.items()}
+            tokens = [lemmatizer.lemmatize(k, pos=v) for k, v in tokens.items()]
+            
+        word_counter = pd.Series(tokens).groupby(tokens).cumcount()
```
tyarkoni: Could also use `pd.Series(tokens).value_counts()`—will produce the same result, but I suspect this version might be slightly faster (though I doubt there's a meaningful difference).

pliers/extractors/text.py
```diff
@@ -402,3 +402,33 @@ def _extract(self, stim):
...
+        
+        
+        return ExtractorResult(list(word_counter), stim,
```
tyarkoni: This approach will probably break because the number of data elements (the first argument) doesn't match the number of features. The expectation is that you're returning a single row (because we have a single `TextStim`), with each feature in a different column. I think what you want here is to pass `features=list(word_counter.index)`.

pliers/extractors/text.py
rbroc: It seems like value_counts returns summary value of number of occurrences for each unique word, rather than the cumulative count for each token in the sequence.

pliers/extractors/text.py
rbroc: should be fixed now

pliers/extractors/text.py
```diff
@@ -402,3 +400,47 @@ def _extract(self, stim):
...
         return ExtractorResult(features_list, stim, self,
                                features=self.features, orders=order_list)
+
+
+class WordCounterExtractor(BatchTransformerMixin, TextExtractor):
+
+    ''' Extracts number of times each unique word has occurred within text
+    
+    Args:
+        lemmatize(bool): specifies if words are to be lemmatized before being counted.
```
adelavega: This probably doesn't need to be here. @rbroc mentioned she wrote a fitler to do this, so that's a better place for it.

pliers/extractors/text.py
```diff
@@ -402,3 +400,47 @@ def _extract(self, stim):
...
                                features=self.features, orders=order_list)
+
+
+class WordCounterExtractor(BatchTransformerMixin, TextExtractor):
+
+    ''' Extracts number of times each unique word has occurred within text
+    
+    Args:
+        lemmatize(bool): specifies if words are to be lemmatized before being counted. 
+        log_scale(bool): specifies if count values are to be returned in log scale (defaults to False)
```
adelavega: I would argue this is also not necessary. The user can easily transform this afterwards.

pliers/extractors/text.py
```diff
@@ -402,3 +400,47 @@ def _extract(self, stim):
...
+        
+        self.lemmatize = lemmatize
+        self.log_scale = log_scale
+        self.features = ['log_word_count'] if self.log_scale else ['word_count']
+        super(WordCounterExtractor, self).__init__()
+        
+    @requires_nltk_corpus
+    def _extract(self, stims):
+        
+        tokens = [s.text.lower() for s in stims]
```
adelavega: Maybe this should be an argument to the Extract init, though? You could potentially want counting to be case sensitive

pliers/extractors/text.py
rbroc: added it because suggested in #340

rbroc:
tests passed, should be pretty much ready for merge or (@adelavega @tyarkoni) another round of review

pliers/extractors/text.py
```diff
@@ -402,3 +400,36 @@ def _extract(self, stim):
...
+    _batch_size = sys.maxsize
+
+    def __init__(self, case_sensitive=False, log_scale=False):
+
+        self.log_scale = log_scale
+        self.case_sensitive = case_sensitive
+        self.features = ['log_word_count'] if self.log_scale else ['word_count']
+        super(WordCounterExtractor, self).__init__()
+
+    def _extract(self, stims):
```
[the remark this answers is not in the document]
rbroc: I wonder whether it makes sense at all to pass `TextStim` to this extractor then. If you get a single TextStim as an input, stim.text can only occur once (as `TextStim` does not tokenize its text).
Wouldn't it make more sense to only allow `ComplexTextStim` as input?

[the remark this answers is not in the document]
tyarkoni: Oh, right, I'd forgotten there was no tokenization. Yeah, I think that's reasonable.

pliers/extractors/text.py
```diff
@@ -402,3 +400,38 @@ def _extract(self, stim):
...
+        tokens = [t if self.case_sensitive else t.lower() for t in tokens]
+        word_counter = pd.Series(tokens).groupby(tokens).cumcount() + 1
+        if self.log_scale:
+            word_counter = np.log(word_counter)
+
+        return ExtractorResult(word_counter, stims, self,
+                               features=self.features,
+                               onsets=onsets, durations=durations)
+
+
```
adelavega:
```suggestion
```

pliers/extractors/text.py
```diff
@@ -402,3 +400,38 @@ def _extract(self, stim):
...
+    def __init__(self, case_sensitive=False, log_scale=False):
+
+        self.log_scale = log_scale
+        self.case_sensitive = case_sensitive
+        self.features = ['log_word_count'] if self.log_scale else ['word_count']
+        super(WordCounterExtractor, self).__init__()
+
+    def _extract(self, stims):
+
```
adelavega:
```suggestion
```

PR #365, opened 2020-01-13
pretrained bert encoding extractor from transformers

rbroc:
Extractor returning encodings from Bert Pretrained models via [transformers](https://github.com/huggingface/transformers). 
Work in progress.


adelavega:
Another minor connect. if `transformers` is `None` (that is, it was not importable bc its not installed), I think a nice error should be thrown. 

This should be more common throughout pliers in fact, as otherwise it gives a cryptic error.

adelavega:
Can you also add `transformers` to optional_dependencies.txt?

pliers/extractors/text.py
```diff
@@ -402,3 +405,131 @@ def _extract(self, stim):
...
+
+        seq, model, tokenizer, pooling = map(expand_metadata,
+                                             [' '.join(text),
+                                              self.pretrained_model,
+                                              self.tokenizer_type,
+                                              str(self.pooling)])
+
+        data = [encodings.tolist(), encoded_tokens, t_text, token_positions,
+                seq, model, tokenizer, pooling]
+        features = ['encoding', 'token', 'word', 'token_position', 'sequence',
```
adelavega: I think some of these should not be features, per se. 

For sure, I think `model`, `tokenizer` and `pooling` should just be attributes of the extractor rather than features. These can always be retrieved from the `Extractor` model. 

Also, shouldn't the pliers built-in `order` be used instead of `token_position`?

Otherwise, the rest seem fine, as they have to do with tokenization done by BERT, which is important.

adelavega:
@tyarkoni this is more of a neuroscout question, but I notice BERT returns a list, in a single field, rather than `n` encoding columns. 

That's fine, but in NS we store all values as strings in the db, so we have to be careful about how we handle this. Would splat-ing this array onto individual columns by pybids' job, or custom neuroscout logic prior to ingestion (fine with that), to ingest as separate features.

pliers/extractors/text.py
```diff
@@ -149,7 +152,7 @@ def __init__(self, variables, missing=np.nan, case_sensitive=True,
             dicts.append(d)
 
         # Make sure none of the dictionaries have duplicate indices
-        drop_dups = lambda d: d[~d.index.duplicated(keep='first')]
+        def drop_dups(d): return d[~d.index.duplicated(keep='first')]
```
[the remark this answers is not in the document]
rbroc: mh... I did not do this. Maybe autopep8? Shall I revert to lambda?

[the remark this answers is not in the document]
rbroc: reverted to lambda as in master branch

pliers/extractors/text.py
rbroc: fixed that in latest commit by dropping those as `features `and adding some `log_attributes` to result df if specified in BERT-specific `to_df` method.
Wondering if it'd be good to add this to generic `to_df` method though (@tyarkoni)

tyarkoni: The generic `to_df` method already includes the `_log_attributes` dictionary as a field, I believe. Beyond that, adding options to `to_df` to control its behavior is complicated by the fact that it's almost always invoked implicitly (from the `ExtractorResult` initializer). So making use of it would require passing through an argument from the `_extract` call in every extractor, which seems clunky.

rbroc: > The generic `to_df` method already includes the `_log_attributes` dictionary as a field, I believe. </br>

I guess the result object does, so definitely always possible to retrieve that info. 

> Beyond that, adding options to `to_df` to control its behavior is complicated by the fact that it's almost always invoked implicitly (from the `ExtractorResult` initializer). So making use of it would require passing through an argument from the `_extract` call in every extractor, which seems clunky. </br>

Ok, still leaving this logic for the bert-specific `to_df` for now though, unless you suggest avoiding to do so here too?

tyarkoni: > Ok, still leaving this logic for the bert-specific to_df for now though, unless you suggest avoiding to do so here too?

If you can already get it from the standard handling, I think you can remove it here. If there's information that's *not* available from the DF otherwise, then leave it in. Either way, I don't think there's a need for the `model_attributes` argument, as I don't think that can currently actually be manipulated by the user via the public API (i.e., you'd have to call `_to_df()` directly, which is discouraged).

rbroc: > If you can already get it from the standard handling, I think you can remove it here. If there's information that's _not_ available from the DF otherwise, then leave it in. Either way, I don't think there's a need for the `model_attributes` argument, as I don't think that can currently actually be manipulated by the user via the public API (i.e., you'd have to call `_to_df()` directly, which is discouraged). </br>

Looks like you _can_ pass `model_attributes` to `to_df` and it will handle them correctly, as `to_df` takes kwargs and passes them to `_to_df()` when calling it internally. 

I can't find a way to plug `log_attributes` into result dataframe from generic `to_df` method though, and find no reference to `log_attributes` where `to_df` is defined. 
Caveat: it might be me not knowing pliers well enough yet.

rbroc:
added torch to optional dependencies, all tests after the one for the new extractor now fail on memory issue ('cannot allocate memory')

pliers/extractors/text.py
```diff
@@ -402,3 +400,138 @@ def _extract(self, stim):
...
+
+        self.pretrained_model = pretrained_model_or_path
+        self.tokenizer_type = tokenizer
+        self.framework = framework
+        self.encoding_level = encoding_level
+        self.pooling = pooling
+        self.model_kwargs = model_kwargs
+        self.tokenizer_kwargs = tokenizer_kwargs
+        model_name = 'BertModel' if self.framework == 'pt' else 'TFBertModel'
+
```
adelavega:
```suggestion
verify_dependencies(['transformers'])
```

I would add a check that transformers was imported. Otherwise you get a cryptic error because `transformers` = None

rbroc:
Added a pytest marker to skip tests on `bert-large-uncased` if tests are running on Travis. 

This test to be the extra step that was making the audio and video tests that come _after_ this trigger the out of memory error.

I've tried out a couple of ablations to diagnose why/how this happens. It was not the `bert-large` test in itself that triggered the memory error, but the combination of torch + tensorflow tests on bert-base-uncased **plus** this. Removing either `bert-large` or tensorflow tests solved the problem and I'd say `bert-large` is lower priority.

Clearing the variables or deleting the models from cache after use did not help. Not sure if Travis caches variables/files also in some other way and if _that_ might be the problem.

Anyway, the tests are passing now and this might be ready for merge. @adelavega @tyarkoni

tyarkoni:
Well if disk space is the issue, this is only going to be a stopgap measure, as we'll likely run into the same problem next time we add a large pretrained model. But I guess we can kick the can down the road for now. Will review now, thanks!

pliers/extractors/text.py
```diff
@@ -402,3 +400,140 @@ def _extract(self, stim):
...
+    ''' Uses transformers library to extract contextualized encodings for words
+    or text sequences using pre-trained BERT models.
+
+    Args:
+        pretrained_model_or_path (str): A string specifying which BERT
+            model to use. Can be one of pretrained BERT models listed at
+            https://huggingface.co/transformers/pretrained_models.html
+            or path to custom model.
+        tokenizer (str): Type of tokenization used in the tokenization step.
```
tyarkoni: What are the valid values here? If this is just passed through, make that explicit and point the reader to the place in the transformers docs that describes them. If it's a small set, probably best to just reproduce them here.

pliers/extractors/text.py
```diff
@@ -402,3 +400,140 @@ def _extract(self, stim):
...
+            If different from model, out-of-vocabulary tokens may be treated as
+            unknown tokens.
+        framework (str): name deep learning framework to use. Must be 'pt'
```
tyarkoni: pedantic, but for consistency, maybe go with either the values "torch" and "tensorflow" or the values "pt" and "tf".

pliers/extractors/text.py
```diff
@@ -402,3 +400,140 @@ def _extract(self, stim):
...
+                 tokenizer='bert-base-uncased',
+                 framework='pt',
+                 encoding_level='token',
+                 pooling=None,
+                 model_kwargs=None,
+                 tokenizer_kwargs=None):
+                 
+        verify_dependencies(['transformers'])
+
+        if framework not in ['pt', 'tf']:
```
tyarkoni: Related to above, `'tf'` is inconsistent with the docstring.

pliers/extractors/text.py
```diff
@@ -402,3 +400,140 @@ def _extract(self, stim):
...
+        self.tokenizer = transformers.BertTokenizer.from_pretrained(
+            tokenizer, **tokenizer_kwargs)
+        self.model = getattr(transformers, model_name).from_pretrained(
+            pretrained_model_or_path, **model_kwargs)
+
+        super(PretrainedBertEncodingExtractor, self).__init__()
+
+    def _extract(self, stims):
+
+        stims.name = ' '.join([s.text for s in stims.elements])
```
tyarkoni: I have mixed feelings about this... I assume the point is to have a string that identifies the `ComplexTextStim`, which makes sense... But the user may have their own reason for setting the `.name` property. Then again, they'd currently have to set that directly, as we don't pass a field through from the `ComplexTextStim` initializer (which is actually a bit of an oversight). So, not sure...

pliers/extractors/text.py
```diff
@@ -402,3 +400,140 @@ def _extract(self, stim):
...
+        text, onsets, durations = zip(
+            *((s.text, s.onset, s.duration) for s in stims.elements))
+        durations = [0.0 if d is None else d for d in durations]
```
tyarkoni: Not sure if we should default to `0` here or `None`. I believe if it's `None` for all values, then the `duration` column would be dropped from the `to_df` call, which might be desirable behavior. Even if that's not true, `0` is different from no value. Is there a reason we need to set this to `0`, as opposed to keeping `None`?

pliers/extractors/text.py
```diff
@@ -402,3 +400,140 @@ def _extract(self, stim):
...
+                                              return_tensors=self.framework)
+        output = self.model(tensor_tokens)
+        output = [out.detach().numpy() if self.framework == 'pt' else
+                  out.numpy() for out in output]
+
+        if self.encoding_level == 'token':
+            encoded_tokens = tokens_flat
+            encodings = output[0][:, 1:-1, :].squeeze()
+
+        elif self.encoding_level == 'sequence':
```
tyarkoni: What happens if `self.encoding_level` isn't either `token` or `sequence`? I assume this is invalid, in which case there should probably be a validation check at the very start of the `__init__` that raises a `ValueError` if `encoding_level not in {'token', 'sequence'}`. Otherwise we could do a lot of loading/computation only to fail here.

pliers/extractors/text.py
rbroc: would adding an "input_sequence" field to log attributes be a viable solution (only returned in result df if model_attributes=True in `_to_df`)?

tyarkoni: `_log_attributes` applies at the transformer instance level, not the stim passed to `transform()`, so I don't think that would work.

Maybe we just bite the bullet and add an `input_sequence` column to the returned DF. This would mess things up a bit in long format, but so be it. (Probably what we should really do is rework the `to_df` logic so that there's a way to tag columns as index vs. content columns.)

tyarkoni: TBH I'm thinking it may have been a bad idea to support the `format` argument in the first place. A lot of the complexities we've faced are introduced by that distinction, and I think it would be pretty reasonable to always return `'wide'` and leave it to users to do any reshaping they need to.

Either way, let's not worry about it in this PR—you can make an executive decision about whether to overwrite `.name` or add a column, and we can revisit the more general issue later.

pliers/extractors/text.py
rbroc: I've specified that BERT models are those preceded by "bert" prefix in the list, it's a pretty long list and might be expanding in the future.

pliers/extractors/text.py
rbroc: fixed (should be tf)

pliers/extractors/text.py
rbroc: fixed

pliers/extractors/text.py
rbroc: the reason was that the logic for the `cast_to_token_level` function included repeating elements using * operator which does not support None objects. Changed the function, replaced * with `itertools.repeat`. Should be fixed now.

pliers/extractors/text.py
rbroc: added `ValueError`

pliers/extractors/text.py
rbroc: ok, sounds like this is a more general problem indeed. I've added a conditional statement which overwrites `.name` only if it is an empty string (leaving some freedom to the user to specify it). It's just a temporary patch, but might do the job until we decide what to do with `to_df`.

pliers/extractors/text.py
```diff
@@ -402,3 +401,147 @@ def _extract(self, stim):
...
+
+        self.tokenizer = transformers.BertTokenizer.from_pretrained(
+            tokenizer, **tokenizer_kwargs)
+        self.model = getattr(transformers, model_name).from_pretrained(
+            pretrained_model_or_path, **model_kwargs)
+
+        super(PretrainedBertEncodingExtractor, self).__init__()
+
+    def _extract(self, stims):
+
```
adelavega:
```suggestion
```

pliers/extractors/text.py
```diff
@@ -402,3 +401,147 @@ def _extract(self, stim):
...
+                encodings = output[1]
+
+        data = [encodings.tolist(), encoded_tokens, t_text]
+        features = ['encoding', 'token', 'word']
+
+        return ExtractorResult(data, stims, self,
+                               features=features,
+                               onsets=t_ons, durations=t_dur)
+
+    def _to_df(self, result, model_attributes=True):
```
adelavega:
```suggestion
    def _to_df(self, result, model_attributes=False):
```

rbroc: Would lean towards leaving default to `True`. I would expect that in most cases you would want to know which token is being encoded.

adelavega: OK, I don't feel strongly about this, especially once my PR is merged.

tyarkoni:
LGTM, but looks like there are merge conflicts after merging one of the other PRs; once they're fixed, this is good to merge.

rbroc:
@tyarkoni should be good now.

PR #366, opened 2020-01-22
Add lemmatization to stemming filter

rbroc:
Adds lemmatization option to `WordStemmingFilter` (via `WordNetLemmatizer` from nltk `stem` module).


pliers/filters/text.py
```diff
@@ -60,14 +63,40 @@ def __init__(self, stemmer='porter', tokenize=True, *args, **kwargs):
...
+    
+        def pos_wordnet(txt):
+            pos_tagged = dict(nltk.pos_tag(txt, tagset='universal'))
+            pos_tagged = {tok: pos_map[tag] if tag in pos_map else 'n' for tok,tag in pos_tagged.items()}
+            return pos_tagged
+        
         if self.tokenize:
-            tokens = stim.text.split()
-            stemmed = ' '.join([self.stemmer.stem(tok) for tok in tokens])
+            tokens = [s.lower() for s in stim.text.split()] if not self.case_sensitive else stim.text.split()
```
tyarkoni: We might want to use nltk's tokenization functionality (minimally, `nltk.word_tokenize`) instead of just splitting on whitespace.

pliers/filters/text.py
```diff
@@ -60,14 +63,40 @@ def __init__(self, stemmer='porter', tokenize=True, *args, **kwargs):
...
+            'NOUN': 'n',
+            'VERB': 'v'
+        }
+    
+        def pos_wordnet(txt):
+            pos_tagged = dict(nltk.pos_tag(txt, tagset='universal'))
+            pos_tagged = {tok: pos_map[tag] if tag in pos_map else 'n' for tok,tag in pos_tagged.items()}
+            return pos_tagged
+        
         if self.tokenize:
```
tyarkoni: I could be missing something, but I think the conditional logic could be simplified. Can we handle the tokenization, lemmatization, and stemming steps serially, instead of branching? At least for the top-level conditional, I think you can just wrap `stim.text` in a list and then the same stemming/lemmatization logic can be applied regardless of tokenization.

pliers/filters/text.py
rbroc: check.

pliers/filters/text.py
rbroc: if I interpreted correctly, this should be fixed by latest commit.

pliers/filters/text.py
```diff
@@ -22,18 +22,20 @@ class TextFilter(Filter):
...
-            or 'rslp'. Alternatively, an initialized nltk StemmerI instance
+            stemming and lemmatization modules available in nltk.stem. 
+            Valid values are 'porter', 'snowball', 'isri', 'lancaster', 'regexp', 
+            'wordnet', or 'rslp'. Alternatively, an initialized nltk StemmerI instance
             can be passed.
-        tokenize (bool): if True, apply the stemmer to each token in the
-            TextStim, otherwise treat the whole TextStim as one token to stem.
+        tokenize (bool): if True, apply the stemmer/lemmatizer to each token in the
+            TextStim, otherwise treat the whole TextStim as one token to stem/lemmatize.
+        case_insensitive (bool): if True, input is lower-cased before stemming
```
tyarkoni: Argument name is `case_insensitive` here, but it's `case_sensitive` in the actual code.

pliers/filters/text.py
```diff
@@ -22,18 +22,20 @@ class TextFilter(Filter):
...
+            or lemmatizing.
         args, kwargs: Optional positional and keyword args passed onto the
-            nltk stemmer.
+            nltk stemmer/lemmatizer.
     '''
 
     stemmers = {
```
tyarkoni: Should probably be `_stemmers`, to discourage access from outside.

pliers/filters/text.py
```diff
@@ -60,14 +65,36 @@ def __init__(self, stemmer='porter', tokenize=True, *args, **kwargs):
...
+            pos_tagged = dict(nltk.pos_tag(txt, tagset='universal'))
+            pos_tagged = {tok: pos_map[tag] if tag in pos_map else 'n'
+                          for tok, tag in pos_tagged.items()}
+            return pos_tagged
+
+        tokens = [stim.text]
         if self.tokenize:
-            tokens = stim.text.split()
+            tokens = nltk.word_tokenize(tokens[0])
+            tokens = [tok if self.case_sensitive else tok.lower()
```
tyarkoni: Maybe de-dent this line so it's outside the conditional? Not sure if lower-casing has any impact in the non-tokenized case, but if it does (or just in case), it should be applied in both cases.

pliers/filters/text.py
rbroc: done

pliers/filters/text.py
rbroc: fixed in latest commit

pliers/filters/text.py
rbroc: fixed in latest commit

pliers/filters/text.py
```diff
@@ -22,52 +23,78 @@ class TextFilter(Filter):
...
+            tokens = nltk.word_tokenize(tokens[0])
+        tokens = [t if self.case_sensitive else t.lower() for t in tokens]
+        if not isinstance(self.stemmer, stem.WordNetLemmatizer):
+            stemmed = ' '.join([self.stemmer.stem(t) for t in tokens])
         else:
-            stemmed = self.stemmer.stem(stim.text)
+            pos_tagged = pos_wordnet(tokens)
+            stemmed = ' '.join([self.stemmer.lemmatize(t, pos=pos_tagged[t])
+                                for t in tokens])
         return TextStim(stim.filename, stemmed)
```
tyarkoni: We probably want to make sure we pass through all of the other fields in the `stim`, so either pass them all explicitly (e.g., `stim.onset`), or just `copy.copy` the stim object and overwrite the text field (the former is probably better).

adelavega:
Looks read to merge to me

tyarkoni:
LGTM once the conflict's resolved.

rbroc:
> LGTM once the conflict's resolved.
will wait until bert_transformer is merged, then rebase this on master to resolve conflicts.

rbroc:
@tyarkoni conflicts resolved.

PR #372, opened 2020-02-04
fix tests

rbroc:
Fixes:
- `SpaCy Extractor`. SpaCy has updated mapping between UD tags ("tag" feature) and the coarser "pos" feature. Mapping for auxiliary updated to UD-tag "VBZ" -> "AUX" instead of "VBZ" -> "VERB". For reference, the tag-to-pos mapping scheme is defined [here](https://github.com/explosion/spaCy/blob/master/spacy/lang/en/morph_rules.py) and the relevant commit where new word-specific rules are defined is [here](https://github.com/explosion/spaCy/commit/04395ffa49409f669147168c8966ac21335cbb4d#diff-452945f82ac41b4b16bd9d7fc658cc1a).
- `PredefinedDictionaryExtractor`: link to Calgary Lexical Decision dataset was broken, now replaced with active link. 
- `_download_dictionary` method edited so to except xlsx files from zipfiles routine.
- Set `pandas` version to < 1.0.0 (should fix API incompatibilities later)
- Drop Python 3.5 from Travis, add 3.7 (Miniconda related)
- Check if `cls` is a class before passing to `issubclass` (for Python 3.7)
- Remove ffmpeg ppa from Travis config (PPA no longer works, used alternative source)
- Edit `TextVectorizerExtractor`. `TfidfVectorizer` from `sklearn.test` is now instance of class `CounterExtractor`, as `VectorizerMixin` class has been dropped in v0.22.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,9 +16,7 @@ matrix:
 before_install:
 - openssl aes-256-cbc -K $encrypted_a0a62c26415d_key -iv $encrypted_a0a62c26415d_iv
   -in pliers/tests/credentials/google.json.enc -out pliers/tests/credentials/google.json -d || true
-- sudo add-apt-repository ppa:mc3man/trusty-media -y
-- sudo apt-get -qq update
-- sudo apt-get install libmp3lame-dev ffmpeg tesseract-ocr graphviz cmake
+- sudo apt-get install libavformat-dev libavfilter-dev libavdevice-dev ffmpeg libmp3lame-dev tesseract-ocr graphviz cmake
 - export MINICONDA=$HOME/miniconda
 - export PATH="$MINICONDA/bin:$PATH"
 - wget $MINICONDA_URL -O miniconda.sh;
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-dist: trusty
+dist: xenial
 language: python
 matrix:
   include:
--- a/optional-dependencies.txt
+++ b/optional-dependencies.txt
@@ -14,7 +14,7 @@ pygraphviz
 pysrt
 pytesseract
 python-twitter
-scikit-learn<=0.21
+scikit-learn
 seaborn
 spacy
 SpeechRecognition>=3.6.0
--- a/optional-dependencies.txt
+++ b/optional-dependencies.txt
@@ -14,7 +14,7 @@ pygraphviz
 pysrt
 pytesseract
 python-twitter
-scikit-learn
+scikit-learn<=0.21
 seaborn
 spacy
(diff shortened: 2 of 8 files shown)


adelavega:
It looks like all but the text extractor issues are due to Python 3.7.

adelavega:
@tyarkoni FYI miniconda dropped support for Python 3.5, so I dropped it from Travis. 
They're going drop 3.6 support soon as well, but we're still OK.

Some changes were needed for 3.7, but relatively minor. 

There's also some pandas 1.0.0 API changes, but they're minor and I sidestepped them for now.

adelavega:
Looks like sklearn > 0.22 causes the `TfidfVectorizer` to only count, and not use TF-IDF. Not sure why, but in the interest of getting tests passing, I just downgraded it.

rbroc:
> Looks like sklearn >= 0.22 causes the `TfidfVectorizer` to only count, and not use TF-IDF. Not sure why, but in the interest of getting tests passing, I just downgraded it.

I've fixed this now, `TfidfVectorizer` works with both latest sklearn and 0.21. 
All tests pass locally, but they fail again on travis with same miniconda error we had earlier (solved when @adelavega forced rebuild)

adelavega:
Awesome. these random failures are really annoying though.

adelavega:
Sweet. The tests finally passed. I had to clear the cache and re-run the tests though. Something to keep in mind for future environment related failures.

tyarkoni:
Is this cumulative with #369? Should we merge that one first?

adelavega:
No, let's merge this one first. I just merged this branch into that one to see if tests passed.

adelavega:
OK, merging this one then.

tyarkoni:
Great, thanks. @rbroc, once this and #369 are merged, do you mind rebasing your other open PRs onto master? Then we can restart the tests before merging them. Thanks!

adelavega:
@tyarkoni FYI one more thing: sometimes tests fail seemingly randomly. I haven't seen it in a while, but it looks like a miniconda issue related to the travis cache. coudn't find anything about it, but clearing the cache seems to work. if it continue to be an issue we could roll back miniconda to a specific version, or just hope that with a new release this problem goes away.

tyarkoni:
I'm fairly certain we've seen this before. Unfortunately I don't remember if there was a solution, if it just went away by itself, or if we just sucked it up and manually cleared the cache occasionally.

adelavega:
It's possible what's going on is that the miniconda dir is cached, and we're trying to install miniconda on top of it. 

I'm going to explore this possible fix: https://github.com/theochem/cgrid/pull/13/files

Issue #370, opened 2020-01-30
Add AudioSet Tensforflow extractors


adelavega:
Google published a dataset called AudioSet, which consists of about 2 million 10-second YouTube clips that are manually annotated on a hierarchical ontology. 
(thanks to @rbroc for finding out about this!)

They then developed and shared TF models for 1) classification of these labels from audio ([Yamnet](https://github.com/tensorflow/models/tree/master/research/audioset/yamnet)) and 2) 128-dimensional embeddings ([VGGish](https://github.com/tensorflow/models/tree/master/research/audioset/vggish)). Both of these models are available with pre-trained networks!

I gave Yamnet a shot on some of our stimuli, and the results are actually pretty good:

Life dataset (60s):
![image](https://user-images.githubusercontent.com/2774448/73477436-d8ff9400-4359-11ea-93b1-210dba27cc61.png)

Sherlock:

![image](https://user-images.githubusercontent.com/2774448/73477487-f7658f80-4359-11ea-8f74-96616e1ffb73.png)

Generating moment by moment labels is quite fast, it should only take a few seconds to produce labels for an entire movie. 

The question is whether this makes sense as a pliers extractor, or something we do outside of pliers, like for facenet. Given that this repo seems fairly well maintained, and extracting for a WAV input is fairly standard, I vote for including in pliers.

The only minor catch is that the audio needs to be sampled at 16000hz. Not sure if we have a filter for downsampling audio, although this is fairly easy to do with various libraries like pydub.

tyarkoni:
Oh, this looks awesome. Agree it should be in pliers and not just in NeuroScout.

tyarkoni:
Oh and we don't have a resampling filter right now (at least for audio) AFAIK. But that would be straightforward to add and I think we should make it its own `AudioResamplingFilter` rather than build it into this one. We could even conceivably add an optional class variable for audio that indicates what sampling rate(s) an `Extractor` needs, but it's probably overkill and the internal logic for these hierarchies is getting pretty convoluted. So maybe best to just raise an exception if a different sampling rate is provided.

rbroc:
closing this as extractor has been implemented

PR #374, opened 2020-02-06
Audio resampling filter

rbroc:
Add `AudioResamplingFilter` based on `librosa.core.resample` as suggested in #370

### DIFF
--- a/pliers/filters/audio.py
+++ b/pliers/filters/audio.py
@@ -34,7 +34,7 @@ class AudioResamplingFilter(AudioFilter):
     
     _log_attributes = ('target_sr', 'resample_type')
     
-    def __init__(self, target_sr=22050, resample_type='kaiser_best',
+    def __init__(self, target_sr=44100, resample_type='kaiser_best',
                  **librosa_kwargs):
         verify_dependencies(['librosa'])
         self.target_sr = target_sr
--- a/pliers/filters/audio.py
+++ b/pliers/filters/audio.py
@@ -1,7 +1,7 @@
 ''' Filters that operate on TextStim inputs. '''
 
 from pliers.stimuli import AudioStim
-from pliers.utils import attempt_to_import
+from pliers.utils import attempt_to_import, verify_dependencies
 from .base import Filter, TemporalTrimmingFilter
 
 librosa = attempt_to_import('librosa')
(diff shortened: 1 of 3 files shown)


rbroc:
tests for python 3.6 pass, while failing on 3.7 (same random error @adelavega mentions #372

pliers/filters/audio.py
```diff
@@ -13,3 +15,39 @@ class AudioFilter(Filter):
...
+            'kaiser_best', 'kaiser_fast', 'scipy', 'fft' or 
+            'polyphase'. See librosa.core.resample documentation for 
+            more details.
+        librosa_kwargs: Optional keyword args passed onto the
+            librosa resampling function.
+    '''
+    
+    _log_attributes = ('target_sr', 'resample_type')
+    
+    def __init__(self, target_sr=22050, resample_type='kaiser_best',
```
tyarkoni: Is there a reason for picking `22050` as the default? Wouldn't 44 KHz be more typical? Or is the assumption that most audio probably already is at >= 44KHz, so if the user's calling this they probably want something lower?

pliers/filters/audio.py
```diff
@@ -13,3 +15,39 @@ class AudioFilter(Filter):
...
+                 **librosa_kwargs):
+        verify_dependencies(['librosa'])
+        self.target_sr = target_sr
+        self.resample_type = resample_type
+        self.librosa_kwargs = librosa_kwargs
+        super(AudioResamplingFilter, self).__init__()
+
+    def _filter(self, stim):
+        stim.data = librosa.core.resample(y=stim.data,
```
tyarkoni: Not sure if librosa's `resample` already does this (doesn't look like it, at a glance), but it might make sense to first check the current sampling rate and only call librosa if the `stim.sampling_rate` is different from the `target_sr`. If the two match, we can just return the existing stim—no need to resample, which is lossy and/or time-consuming.

pliers/filters/audio.py
rbroc: I was indeed assuming that 44.1 KHz would be more typical. Leaving it as it is unless you're strongly against it.

tyarkoni: I don't feel strongly about it, but I guess I lean towards `44100`. The reasoning is mainly that I expect this filter to be used mostly in the service of preparing audio for some other API or service that insists on a particular sampling rate. My naive guess is that the most common such rate would be 44.1 KHz. But I'm fine leaving it as is.

rbroc: default changed to 44100

tyarkoni:
@rbroc is this ready to merge?

rbroc:
> @rbroc is this ready to merge?

yes, unless we want to wait for tests to pass.

PR #375, opened 2020-02-06
Try conditional caching of miniconda

adelavega:
Trying to fix random miniconda failures, inspired by: https://github.com/theochem/cgrid/blob/03a995a602fe043ee9c05e3797ffaa9836fa031f/.travis.yml


tyarkoni:
LGTM (I'm fine merging this change, whether or not it solves the actual problem)

adelavega:
Cool. I restarted the tests one final time, to test if it uses the cache properly. I tested without cache already and it worked. So if this passes I'm pretty sure it fixes the problem and I'll merge.

adelavega:
Looks like its working. Travis failure seems to be something else (timed out, but after env set up)

PR #376, opened 2020-02-07
Fix index_cols in to_df

adelavega:
- First off, a bug fix. If `object_id` is in df.columns (in `to_df`), `object_id` should be in the `index_cols` otherwise it will be melted as a features if you call `.to_df(format='long')`. 
- Second, @rbroc and I have been going back on forth on whether `index_cols` should actually be not only the reserved columns (e.g. `onset`, `duration` `object_id`), but also any non-features. 

For example, in BERT's `_to_df` function, if you say `to_df(format='long', model_attributes=True)`, the model_attributes (such as `tokenizer_type` and `pooling`), will be melted as features into long format. 

I've been making the argument that `to_df` currently is only reserved keys + features, but @rbroc has argued that extra columns would be useful to include stimulus level attributes that are not "features".

@tyarkoni ?


adelavega:
No idea why, but tests are failing on Python 3.6 due to a timeout.

adelavega:
Looks like we need to make a decision about whether `index_cols` should be a) specific special keys such as: `object_id` + `onset` `duration` etc, b) all columns except `ext.features` or c )  the keys in (a) + `ext._log_attributes`.

Any opinion here? My preference is a) or c).

tyarkoni:
I'd probably go with (c).

pliers/extractors/base.py
```diff
@@ -110,8 +110,13 @@ def to_df(self, timing=True, metadata=False, format='wide',
         # Ideally, Extractors should implement their own _to_df() class method
         # that produces a DataFrame in standardized format. Failing that, we
         # assume self._data is already array-like and can be wrapped in a DF.
+        index_cols = []
+
         if hasattr(self.extractor, '_to_df'):
             df = self.extractor._to_df(self, **to_df_kwargs)
+            for col in self.extractor._log_attributes:
```
tyarkoni: I could be remembering wrong, but isn't the dict in `_log_attributes` stored as a string in a *single* column in the default (wide) format? If that's the case, then I think we should respect that in long format and keep storing them all in one field as a string. If I'm wrong and they're already represented as individual fields, then this is fine.

adelavega: I just tried it out with  another extractor (`PredefinedDictionaryExtractor`), and it look like by default you don't get log attributes back in `to_df` at all. 

So this is really just to allow specific extractors to if they so wish, without screwing up the long format.

adelavega: I'd also be fine without this all together, but @rbroc wanted a way to pass back tokenizer results for BERT (which IMO are not actual "features") in `to_df`,

adelavega:
In the end, I decided to go with b) all column except features.

I really can't think of a scenario when this would be the wrong thing to do, and this frees up extractors to do whatever they want in `_to_df` and still be compatible with long format.

adelavega:
Damn tests are timing out again, but they passed once by chance, so once everyone agrees this approaches makes sense, we can merge (so that the BERT branch is also mergable)

rbroc:
> Damn tests are timing out again, but they passed once by chance, so once everyone agrees this approaches makes sense, we can merge (so that the BERT branch is also mergable)

makes sense to me!

PR #378, opened 2020-02-13
Skip test if in travis

adelavega:
- Try skipping test that is timing out on travis


rbroc:
the reason why the second test in` test_io.py `and `test_remote_stims` in `test_stims.py` fail is a server error with http://www.obamadownloads.com/ from which we're trying to download mp4 files.
Replacing this with new video should fix that.

adelavega:
@tyarkoni for some magical random reason tests passed. looks like the number of tests is just adding up memory wise, as we end up w/ about 1.75GB of memory used. This should be OK, but who knows what is actually going on in travis, as tests sometimes pass then sometimes dont.

The best we could do is skip tests that use up a lot of memory consistently on travis only, and hope most of the time that works...

PR #379, opened 2020-02-17
yamnet architecture for audioset label extractor

rbroc:
wrapper for https://github.com/tensorflow/models/blob/master/research/audioset/yamnet/yamnet.py extracting label probability for 521 audio events from AudioSet hierarchical ontology.

### DIFF
--- a/pliers/extractors/__init__.py
+++ b/pliers/extractors/__init__.py
@@ -137,5 +137,6 @@ __all__ = [
     'TempoExtractor',
     'BeatTrackExtractor',
     'HarmonicExtractor',
-    'PercussiveExtractor'
+    'PercussiveExtractor',
+    'AudiosetLabelExtractor'
 ]
--- a/pliers/extractors/audio.py
+++ b/pliers/extractors/audio.py
@@ -530,7 +530,7 @@ class AudiosetLabelExtractor(AudioExtractor):
             yamnet_path = YAMNET_PATH
         try:
             sys.path.insert(0, str(yamnet_path))
-            yamnet = attempt_to_import('yamnet')
+            self.yamnet = attempt_to_import('yamnet')
             verify_dependencies(['yamnet'])
         except MissingDependencyError: 
             msg = ('Yamnet could not be imported. To download and set up '
@@ -538,23 +538,23 @@ class AudiosetLabelExtractor(AudioExtractor):
             raise MissingDependencyError(dependencies=None,
                                          custom_message=msg)
         verify_dependencies(['tensorflow'])
-        
         if top_n and labels:
             raise ValueError('Top_n and labels are mutually exclusive '
                              'arguments. Reinstantiate the extractor setting '
                              'top_n or labels to None (or leaving it '
                              'unspecified).')
 
-        MODULE_PATH = path.dirname(yamnet.__file__)
+        MODULE_PATH = path.dirname(self.yamnet.__file__)
         LABELS_PATH = path.join(MODULE_PATH, 'yamnet_class_map.csv')
         self.weights_path = weights_path or path.join(MODULE_PATH, 'yamnet.h5')
         self.hop_size = hop_size
         self.yamnet_kwargs = yamnet_kwargs or {}
-        self.params = yamnet.params.__dict__
-        self.params = {k: v for k, v in self.params.items() if k.isupper()}
-        self.params['PATCH_HOP_SECONDS'] = hop_size
-        self.params.update(self.yamnet_kwargs)
-        if self.params['PATCH_WINDOW_SECONDS'] != 0.96:
+        self.params = self.yamnet.params
+        self.params.PATCH_HOP_SECONDS = hop_size
+        for par, v in self.yamnet_kwargs.items():
+            if par in self.params.__dict__:
+                setattr(self.params, par, v)
+        if self.params.PATCH_WINDOW_SECONDS != 0.96:
             logging.warning('Custom values for PATCH_WINDOW_SECONDS were '
                 'passed. YAMNet was trained on windows of 0.96s. Different '
                 'values might yield unreliable results.')
@@ -575,31 +575,31 @@ class AudiosetLabelExtractor(AudioExtractor):
         super(AudiosetLabelExtractor, self).__init__()
 
     def _extract(self, stim):
-        params = self.params
-        params['SAMPLE_RATE'] = stim.sampling_rate
+        self.params.SAMPLE_RATE = stim.sampling_rate
 
-        if params['SAMPLE_RATE'] >= 2 * params['MEL_MAX_HZ']:
-            if params['SAMPLE_RATE'] != 16000:
+        if self.params.SAMPLE_RATE >= 2 * self.params.MEL_MAX_HZ:
+            if self.params.SAMPLE_RATE != 16000:
                 logging.warning(
                     'The sampling rate of the stimulus is '
-                    '{}Hz. YAMNet was trained on audio sampled at 16000Hz. ' 
-                    'This should not impact predictions, but you can resample ' 
-                    'the input using AudioResamplingFilter for full conformity ' 
-                    'to training.'.format(params['SAMPLE_RATE']))
-            if params['MEL_MIN_HZ'] != 125 or params['MEL_MAX_HZ'] != 7500:
+                    f'{self.params.SAMPLE_RATE}Hz. YAMNet was trained on '
+                    ' audio sampled at 16000Hz. This should not impact '
+                    'predictions, but you can resample the input using '
+                    'AudioResamplingFilter for full conformity ' 
+                    'to training.')
+            if self.params.MEL_MIN_HZ != 125 or self.params.MEL_MAX_HZ != 7500:
                 logging.warning(
                     'Custom values for MEL_MIN_HZ and MEL_MAX_HZ '
-                    'were passed. Changing these defaults might affect ' 
+                    'were passed. Changing these defaults might affect '
                     'model performance.')
         else:
             raise ValueError(
-                'The sampling rate of your stimulus ({}Hz) must be at least '
-                'twice the value of MEL_MAX_HZ ({}Hz). Upsample your audio '
-                'stimulus (recommended) or pass a lower value of MEL_MAX_HZ '
-                'when initializing the extractor'
-                '.'.format(params['SAMPLE_RATE'], params['MEL_MAX_HZ']))
+                'The sampling rate of your stimulus '
+                f'({self.params.SAMPLE_RATE}Hz) must be at least twice the '
+                f'value of MEL_MAX_HZ ({self.params.MEL_MAX_HZ}Hz). Upsample'
+                ' your audio stimulus (recommended) or pass a lower value of '
+                'MEL_MAX_HZ when initializing the extractor.')
 
-        model = yamnet.yamnet_frames_model(params)
+        model = self.yamnet.yamnet_frames_model(self.params)
         model.load_weights(self.weights_path)
         preds, _ = model.predict_on_batch(np.reshape(stim.data, [1,-1]))
         preds = preds.numpy()[:,self.label_idx]
@@ -609,12 +609,12 @@ class AudiosetLabelExtractor(AudioExtractor):
         preds = np.fliplr(preds[:,idx][:,-nr_lab:])
         labels = [self.labels[i] for i in idx][-nr_lab:][::-1]
 
-        hop = params['PATCH_HOP_SECONDS']
-        window = params['PATCH_WINDOW_SECONDS']
-        stft_hop = params['STFT_HOP_SECONDS']
-        stft_window = params['STFT_WINDOW_SECONDS']
+        hop = self.params.PATCH_HOP_SECONDS
+        window = self.params.PATCH_WINDOW_SECONDS
+        stft_window = self.params.STFT_WINDOW_SECONDS
+        stft_hop = self.params.STFT_HOP_SECONDS
         dur = window + stft_window - stft_hop
-        onsets = np.arange(start=0, stop=stim.duration + dur, step=hop)
+        onsets = np.arange(start=0, stop=stim.duration - dur, step=hop)
 
         return ExtractorResult(preds, stim, self, features=labels,
                                onsets=onsets, durations=dur,
--- a/pliers/extractors/audio.py
+++ b/pliers/extractors/audio.py
@@ -513,7 +513,7 @@ class AudiosetLabelExtractor(AudioExtractor):
         parameters for the model (see params.py file in yamnet repository)
     '''
 
-    _log_attributes = ('yamnet_kwargs')
+    _log_attributes = ()
 
     def __init__(self, hop_size=0.1, top_n=None, 
                  spectrogram=False, **yamnet_kwargs):
@@ -541,16 +541,19 @@ class AudiosetLabelExtractor(AudioExtractor):
         params = self.params
         params.SAMPLE_RATE = stim.sampling_rate
         
-        with tf.Graph().as_default():
+        tf_graph = tf.Graph()
+        with tf_graph.as_default():
             model = yamnet.yamnet_frames_model(params)
             model.load_weights(self.weights_path)
             preds, spectrogram = model.predict(np.reshape(data, [1,-1]), 
                                                   steps=1)
         self.spectrogram = spectrogram
         
+        ########## CHECK HERE ########## 
         idx = np.mean(preds,axis=0).argsort()
         preds = np.flip(preds[:,idx],axis=0)[:,:self.top_n]
         labels = [self.labels[i] for i in idx][::-1][:self.top_n]
+        ################################
 
         durations = params.PATCH_HOP_SECONDS
         onsets = np.arange(start=0, stop=stim.duration, step=durations)
@@ -560,9 +563,8 @@ class AudiosetLabelExtractor(AudioExtractor):
                                onsets=onsets, durations=durations,
                                orders=list(range(len(onsets))))
 
+# Check model values
 # Add pointer to installation instructions (install models, run test install - link, add to pythonpath)
-# Add length warning
-# Migrate to models?
-# Plot predictions?
-# Log sampling rate and kwargs
-# issue with reinitializing after call
+# Add length warning / batching
+# Migrate to models
+# Look into kwargs log
--- a/pliers/extractors/audio.py
+++ b/pliers/extractors/audio.py
@@ -561,14 +561,25 @@ class AudiosetLabelExtractor(AudioExtractor):
         for par, v in self.yamnet_kwargs.items():
             if par in self.params.__dict__:
                 setattr(self.params, par, v)
-        self.labels = pd.read_csv(LABELS_PATH)['display_name'].tolist()
-        self.label_subset = label_subset
+        
+        if top_n and label_subset:
+            raise ValueError('Only one of top_n or label_subset can be provided. '
+                             'Reinitialize the extractor setting either top_n or '
+                             'subset_label to None (or leaving it unspecified)')
+                             
         self.top_n = top_n
-        if self.label_subset:
-            for l in self.label_subset:
-                if l not in self.labels:
-                    logging.warning('''Label {} does not exist. 
-                                     Dropping.'''.format(l))
+        all_labels = pd.read_csv(LABELS_PATH)['display_name'].tolist()
+        if label_subset:
+            for l in label_subset:
+                if l not in all_labels:
+                    logging.warning(f'Label {l} does not exist. Dropping.')
+                    label_subset.remove(l)
+            self.labels = label_subset
+            self.label_idx = [idx for idx, lab in enumerate(all_labels) 
+                                    if lab in label_subset]
+        else:
+            self.labels = all_labels
+            self.label_idx = range(len(all_labels))
         super(AudiosetLabelExtractor, self).__init__()
 
     def _extract(self, stim):
@@ -597,27 +608,15 @@ class AudiosetLabelExtractor(AudioExtractor):
                 'Upsample your audio stimulus (recommended) or pass a lower '
                 'value of MEL_MAX_HZ when initializing this extractor.')
 
-        labels = self.labels
         model = yamnet.yamnet_frames_model(params)
         model.load_weights(self.weights_path)
         preds, _ = model.predict_on_batch(np.reshape(stim.data, [1,-1]))
-        preds = preds.numpy()
-
-        if self.label_subset:
-            label_subset_idx = [idx for idx, lab in enumerate(labels) 
-                                    if lab in self.label_subset]
-            preds = preds[:,label_subset_idx]
-            labels = self.label_subset
-
-        nr_lab = self.top_n or len(labels)
-        if nr_lab > len(labels):
-            raise ValueError(
-                f'Value of top_n ({self.top_n}) exceeds number of '
-                f'labels ({len(labels)}). Reinstantiate this extractor using '
-                'suitable parameters.''')
+        preds = preds.numpy()[:,self.label_idx]
+        
+        nr_lab = self.top_n or len(self.labels)
         idx = np.mean(preds,axis=0).argsort()
         preds = np.fliplr(preds[:,idx][:,-nr_lab:])
-        labels = [labels[i] for i in idx][-nr_lab:][::-1]
+        labels = [self.labels[i] for i in idx][-nr_lab:][::-1]
 
         durations = params.PATCH_HOP_SECONDS
         onsets = np.arange(start=0, stop=stim.duration, step=durations)
@@ -626,5 +625,3 @@ class AudiosetLabelExtractor(AudioExtractor):
         return ExtractorResult(preds, stim, self, features=labels,
                                onsets=onsets, durations=durations,
                                orders=list(range(len(onsets))))
-
-# Add pointer to installation instructions (install models, run test install - link, add to pythonpath)
--- a/pliers/extractors/audio.py
+++ b/pliers/extractors/audio.py
@@ -9,6 +9,7 @@ import pandas as pd
 import soundfile as sf
 from abc import ABCMeta
 from os import path
+import warnings
 
 librosa = attempt_to_import('librosa')
 yamnet = attempt_to_import('yamnet')
@@ -493,64 +494,62 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
     For details on argument specification visit:
     https://librosa.github.io/librosa/effect.html.'''
 
+
     _feature = 'percussive'
 
 
 class AudiosetLabelExtractor(AudioExtractor):
 
     ''' Extract probability of audio event classes based on AudioSet-YouTube 
-    corpus using a YAMNET architecture. 
-    Code available at:
+    corpus using a YAMNet architecture. Code available at:
     https://github.com/tensorflow/models/tree/master/research/audioset/yamnet 
 
     Args:
-    hop_size (float): size of the window (in seconds) on which label extraction 
-                      is performed.
-    sample_rate (int): sampling rate parameter for the yamnet model. Must match
-                        stimulus sampling rate. Note that yamnet was trained
-                        on waveforms resampled at 16000Hz.
-    top_n (int or False): defines how many of the highest label probabilities 
-                          are returned.
+    hop_size (float): size of the window (in seconds) on which label extraction
+        is performed.
+    top_n (int): if defined, specifies how many of the highest label 
+        probabilities are returned.
     spectrogram (bool): if True, plots mel-spectrogram of the audio stimulus.
+    yamnet_kwargs (optional): Optional named arguments that modify input 
+        parameters for the model (see params.py file in yamnet repository)
     '''
-    _log_attributes = ()
 
-    def __init__(self, hop_size=0.1, sample_rate=16000, top_n=False, 
+    _log_attributes = ('params', 'top_n')
+
+    def __init__(self, hop_size=0.1, top_n=None, 
                  spectrogram=False, **yamnet_kwargs):
         verify_dependencies(['yamnet'])
         verify_dependencies(['tensorflow'])
-
-        self.params = yamnet.params
-        self.params.PATCH_HOP_SECONDS = hop_size
-        self.params.SAMPLE_RATE = sample_rate
-        self.spectrogram = spectrogram
-        self.top_n = top_n
-        self.tf_graph = tf.Graph()
 
         MODULE_PATH = path.dirname(yamnet.__file__)
         WEIGHTS_PATH = path.join(MODULE_PATH, 'yamnet.h5')
         LABELS_PATH = path.join(MODULE_PATH, 'yamnet_class_map.csv')
 
+        self.spectrogram = spectrogram
+        self.params = yamnet.params
+        self.params.PATCH_HOP_SECONDS = hop_size
+
         if yamnet_kwargs:
-            for p in self.params.__dict__:
-                if p in yamnet_kwargs.keys():
-                    setattr(self.params, p, yamnet_kwargs[p])
+            for par in self.params.__dict__:
+                if yamnet_kwargs[par]:
+                    setattr(self.params, par, yamnet_kwargs[par])
+
+        self.tf_graph = tf.Graph()
         with self.tf_graph.as_default():
             self.model = yamnet.yamnet_frames_model(self.params)
             self.model.load_weights(WEIGHTS_PATH)
+
         self.labels = pd.read_csv(LABELS_PATH)['display_name'].tolist()
+        self.params = {par: val for par,val in self.params.__dict__.items()
+                       if par.isupper()}
+        self.top_n = top_n if top_n else len(self.labels)
         super(AudiosetLabelExtractor, self).__init__()
 
     def _extract(self, stim):
-        try:
-            assert stim.sampling_rate == self.params.SAMPLE_RATE
-        except:
-            raise ValueError('''Stimulus sampling rate does not match
-                model sampling rate. Resample your stimulus using
-                AudioResamplingFilter or reinitialize this extractor
-                with correct sample_rate value''')
-
+        
         data = stim.data
+        self.params['SAMPLE_RATE'] = stim.sampling_rate
+        
         with self.tf_graph.as_default():
             preds, spectrogram = self.model.predict(np.reshape(data, [1,-1]), 
                                                     steps=1)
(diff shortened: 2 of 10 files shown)


rbroc:
@tyarkoni @adelavega the basic logic should be in place and working now, could you take a quick first look to see if you have any major objections? 

Stlill missing:
- tests
- docs on how to download and setup yamnet; 
- test long audio files and add warning if file is too long/heavy;
- add option to display heatmap with probability over time 

Also wondering if this should be an audio extractor or in models.

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+                 spectrogram=False, **yamnet_kwargs):
+        verify_dependencies(['yamnet'])
+        verify_dependencies(['tensorflow'])
+
+        MODULE_PATH = path.dirname(yamnet.__file__)
+        WEIGHTS_PATH = path.join(MODULE_PATH, 'yamnet.h5')
+        LABELS_PATH = path.join(MODULE_PATH, 'yamnet_class_map.csv')
+
+        self.spectrogram = spectrogram
+        self.params = yamnet.params
```
adelavega: Maybe only the kwarg params should be logged. That is, only log if you changed from the defaults

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+                if yamnet_kwargs[par]:
+                    setattr(self.params, par, yamnet_kwargs[par])
+
+        self.tf_graph = tf.Graph()
+        with self.tf_graph.as_default():
+            self.model = yamnet.yamnet_frames_model(self.params)
+            self.model.load_weights(WEIGHTS_PATH)
+
+        self.labels = pd.read_csv(LABELS_PATH)['display_name'].tolist()
+        self.params = {par: val for par,val in self.params.__dict__.items()
```
adelavega: I would do this earlier, when checking `kwargs`. I can't see another place where non-upper case args would get in. Especially if users can only modify arguments that are aleady in `params.py` (that is, kwargs only get applied if they match an existing key, over-riding a default).

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+                       if par.isupper()}
+        self.top_n = top_n if top_n else len(self.labels)
+        super(AudiosetLabelExtractor, self).__init__()
+
+    def _extract(self, stim):
+
```
adelavega:
```suggestion
```

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+        self.labels = pd.read_csv(LABELS_PATH)['display_name'].tolist()
+        self.params = {par: val for par,val in self.params.__dict__.items()
+                       if par.isupper()}
+        self.top_n = top_n if top_n else len(self.labels)
+        super(AudiosetLabelExtractor, self).__init__()
+
+    def _extract(self, stim):
+        
+        data = stim.data
+        self.params['SAMPLE_RATE'] = stim.sampling_rate
```
adelavega: Are you sure its OK to do this? You explained to me that it was only change how the spectrogram is created, so if you're positive, sounds good to me. I'm just noting this comment in params.py:


> The following hyperparameters (except PATCH_HOP_SECONDS) were used to train YAMNet,
> so expect some variability in performance if you change these. The patch hop can
> be changed arbitrarily: a smaller hop should give you more patches from the same
> clip and possibly better performance at a larger computational cost.

adelavega: Might be worth testing the same audio clip with two sampling rates

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+
+        durations = self.params['PATCH_HOP_SECONDS']
+        onsets = np.arange(start=0, stop=stim.duration, step=durations)
+        onsets = onsets[onsets + self.params['PATCH_WINDOW_SECONDS'] < stim.duration]
+
+        return ExtractorResult(preds, stim, self, features=self.labels,
+                               onsets=onsets, durations=durations,
+                               orders=list(range(len(onsets))))
+
+# Add pointer to installation instructions (install models, run test install - link, add to pythonpath)
```
adelavega: What will this look like? It's not pip installable right?

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+# Add length warning
+# Migrate to models?
```
adelavega: I could see it living in either place. Maybe models makes more sense, but it can still import `AudioExtractor` and subclass from it.

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+    def _extract(self, stim):
+        
+        data = stim.data
+        self.params['SAMPLE_RATE'] = stim.sampling_rate
+        
+        with self.tf_graph.as_default():
+            preds, spectrogram = self.model.predict(np.reshape(data, [1,-1]), 
+                                                    steps=1)
+        
+        if self.spectrogram:
```
adelavega: Yeah, don't think this belong here. Instead, the user is free to inspect the `Extractor` and inspect/plt the spectrogram.

This could maybe be a misc tool, but its easy enough to do I'm not too worried.

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+        durations = self.params['PATCH_HOP_SECONDS']
+        onsets = np.arange(start=0, stop=stim.duration, step=durations)
+        onsets = onsets[onsets + self.params['PATCH_WINDOW_SECONDS'] < stim.duration]
+
+        return ExtractorResult(preds, stim, self, features=self.labels,
+                               onsets=onsets, durations=durations,
+                               orders=list(range(len(onsets))))
+
+# Add pointer to installation instructions (install models, run test install - link, add to pythonpath)
+# Add length warning
```
adelavega: We discussed this, but if there was a mixin to batch the input, that would be great. Or if it was just done within here. As a user, it seems annoying to have to manually filter, especially since the results can be stiched back together after batching.

adelavega:
Oh, and this works in TF 1 & 2 right?

pliers/extractors/audio.py
rbroc: nope, I guess it'd have to be part of the docstring pointing to the instructions on the yamnet repo + info on how to add yamnet path to PYTHONPATH.

pliers/extractors/audio.py
rbroc: ok, let's wait with this and hear what @tyarkoni's take is, the extractor can be easily migrated later on

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+    ''' Extract probability of audio event classes based on AudioSet-YouTube 
+    corpus using a YAMNet architecture. Code available at:
+    https://github.com/tensorflow/models/tree/master/research/audioset/yamnet 
+
+    Args:
+    hop_size (float): size of the window (in seconds) on which label extraction
+        is performed.
+    top_n (int): specifies how many of the highest label probabilities are 
+        returned. If not defined, returns probabilities for all labels.
+    spectrogram (bool): if True, plots mel-spectrogram of the audio stimulus.
```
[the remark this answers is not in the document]
rbroc: added spectrogram as field to Extractor, so that user can retrieve it from there.

pliers/extractors/audio.py
rbroc: removed plotting option and flag, now spectrogram is retrievable from Extractor object.

pliers/extractors/audio.py
rbroc: edited log_attributes as suggested

pliers/extractors/audio.py
rbroc: unfortunately `yamnet_frames_model` internally calls specific parameters from `params` as  `params.PARAM_NAME`, which is why `params` is converted to dictionary only here. I don't see any easy workaround, but open to suggestions.

pliers/extractors/audio.py
rbroc: > Oh, and this works in TF 1 & 2 right?

yes, found a workaround to `yamnet` versioning issue by explicitly calling tf graph. With will keep ensuring that it's compatible with both. Also raised the issue on yamnet repo (https://github.com/tensorflow/models/issues/8157), something might happen on that front too.

rbroc:
> Oh, and this works in TF 1 & 2 right?
Yes, found a workaround for the yamnet versioning issue by calling tf graph explicitly, which works both in tf 1 and 2. Will keep ensuring that it's compatible with both. Also raised an issue on yamnet repo, something might happen on that front too (https://github.com/tensorflow/models/issues/8157).

pliers/extractors/audio.py
```diff
@@ -488,4 +494,88 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+class AudiosetLabelExtractor(AudioExtractor):
+
+    ''' Extract probability of audio event classes based on AudioSet-YouTube 
+    corpus using a YAMNet architecture. Code available at:
+    https://github.com/tensorflow/models/tree/master/research/audioset/yamnet 
+
+    Args:
+    hop_size (float): size of the window (in seconds) on which label extraction
+        is performed.
+    top_n (int): specifies how many of the highest label probabilities are
```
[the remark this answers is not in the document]
rbroc: will enable option to subset output to only include a list of categories to start with, then we can take hierarchy issue from there :)

pliers/extractors/audio.py
rbroc: will make sure to test it thoroughly (and look into an issue with output values I just noticed) and call another round of review :)

rbroc:
next to dos:
- look into sampling rate issue
- look into scale of output values
- batching

pliers/extractors/audio.py
rbroc: okay, I've looked into this and I think I'm on top of it.
Tl;dr -> it's okay to have other `SAMPLE_RATE` than 16000Hz, as a) size of input patches will still be the same; b) results should not be significantly affected. Caveat is that `SAMPLE_RATE` has to be twice as large as `MEL_MAX_HZ`. Writing down a lengthy explanation below, mostly as a reminder for myself. :)

### (Why) the model is compatible with different sampling rates
Yamnet is compatible with different sampling rates, although it's been trained with audio sampled at 16000Hz. 

The input sampling rate is only relevant in the preprocessing step, where the waveform is first passed through STFT to extract spectrogram, which is then converted into a mel-scale spectrogram. 

The sampling rate parameter is used to compute how many samples in the stimulus go into a STFT window, but the size of the window is independent from `SAMPLE_RATE` and explicitly encoded in a separate parameter. Therefore, different sampling rates do not affect how the waveform is binned in the **temporal** domain. 

What `SAMPLE_RATE` does implicitly influence, though, is the (number, values and resolution of) **frequency bins** used for the Fourier transform. Higher `SAMPLE_RATE` correspond to higher frequency resolution in the spectrogram. 

The second pass is transforming the spectrogram into a mel-scale spectrogram. The trick here is that, regardless of how many and which frequency bins you have in the spectrogram, the number of frequency bands in the mel spectrogram is fixed by the parameters `MEL_BANDS` (= number of mel-frequency bins, set to 64 for training) and its range is determined by `MEL_MIN_HZ` / `MEL_MAX_HZ` (set to 125/7500 Hz for training). Input spectrograms of size [n_freq_bins, n_time_bins] will always output [n_mel_bands, n_time_bins], where n_mel_bands is independent of n_freq_bins. That's why different sampling rates would yield still fit model requirements in terms of input size. 

### Different sampling rates should have no impact on the results
Although the input shape is not affected, having different sampling rates on the same waveform _might_ still yield slightly differences in the spectrograms (as FFT frequency resolution will differ).
But as long as the mel frequency range (`MEL_MIN_HZ` and `MEL_MAX_HZ`) is kept close to that used for training these differences shouldn't have any impact. 

### Which of the params can be changed?
- `MEL_BANDS` has to be 64 as the model expects this in its input
- `MEL_MIN_HZ` and `MEL_MAX_HZ` can be changed, with the caveat that model performance can actually be affected and that `MEL_MAX_HZ` has to be less than half of the sampling rate (Nyquist freq).
- `SAMPLE_RATE` can be modified and adapted to actual sampling rate, with the constraint that it has to be at least twice as large as` MEL_MAX_HZ`.

### Re: pliers extractor
- Would probably add a warning if `SAMPLE_RATE` ≠ 16000Hz telling the user that the model was trained on 1600Hz, different sampling rates may have a minimal impact on the results and it could be an idea to resample;
- Would also add a pliers error if `SAMPLE_RATE` < 2 * `MEL_MAX_HZ`, warning that it may be a bad idea to have lower sampling rate and suggesting to either upsample or change `MEL_MAX_HZ` to a suitable value, although this may reduce accuracy of the predictions

rbroc:
@adelavega @tyarkoni this is ready for review.
All the functionality of the extractor should be now in place. 
I've also added tests (passing), changed the travis configuration file to set up the yamnet module, and fixed an issue with the audio resampling filter.

pliers/extractors/audio.py
```diff
@@ -488,4 +494,140 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+                    'model performance.')
+        else:
+            raise ValueError(
+                f'The sampling rate of your stimulus ({params.SAMPLE_RATE}Hz)'
+                ' must be at least twice the value of MEL_MAX_HZ '
+                f'({params.MEL_MAX_HZ}Hz). ' 
+                'Upsample your audio stimulus (recommended) or pass a lower '
+                'value of MEL_MAX_HZ when initializing this extractor.')
+
+        labels = self.labels
```
adelavega: Is this necessary? it refers to the same variable anyways, so maybe just refer to `self.labels` further down.

pliers/extractors/audio.py
```diff
@@ -488,4 +494,140 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+        model = yamnet.yamnet_frames_model(params)
+        model.load_weights(self.weights_path)
+        preds, _ = model.predict_on_batch(np.reshape(stim.data, [1,-1]))
+        preds = preds.numpy()
+
+        if self.label_subset:
+            for l in self.label_subset:
+                print(l)
```
adelavega:
```suggestion
```
Remove print statements

pliers/extractors/audio.py
```diff
@@ -488,4 +494,140 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+        LABELS_PATH = path.join(MODULE_PATH, 'yamnet_class_map.csv')
+        self.weights_path = weights_path or path.join(MODULE_PATH, 'yamnet.h5')
+        self.hop_size = hop_size
+        self.yamnet_kwargs = yamnet_kwargs
+        self.params = yamnet.params
+        self.params.PATCH_HOP_SECONDS = hop_size
+        for par, v in self.yamnet_kwargs.items():
+            if par in self.params.__dict__:
+                setattr(self.params, par, v)
+        self.labels = pd.read_csv(LABELS_PATH)['display_name'].tolist()
```
adelavega: maybe call this `self.all_labels` to differentiate from the labels selected in the end.

pliers/extractors/audio.py
```diff
@@ -488,4 +494,140 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+        model = yamnet.yamnet_frames_model(params)
+        model.load_weights(self.weights_path)
+        preds, _ = model.predict_on_batch(np.reshape(stim.data, [1,-1]))
+        preds = preds.numpy()
+
+        if self.label_subset:
+            for l in self.label_subset:
+                print(l)
+                print(np.where(labels == l))
+            label_subset_idx = [idx for idx, lab in enumerate(labels)
```
adelavega: The logic of which labels are kept and how the indexes are handled gets a bit confusing.

I would try to handle as much of this upfront in the `__init__` function. I would make `top_n` and `labels_subset` mutually exclusive arguments, such that specifying both raises an error. 

Then, compute the idxs up front for the subset_labels, and for `top_n` I suppose you have to compute that in `_extract` as it depends on the results.

pliers/extractors/audio.py
```diff
@@ -488,4 +494,140 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+            for l in self.label_subset:
+                print(l)
+                print(np.where(labels == l))
+            label_subset_idx = [idx for idx, lab in enumerate(labels) 
+                                    if lab in self.label_subset]
+            preds = preds[:,label_subset_idx]
+            labels = self.label_subset
+
+        nr_lab = self.top_n or len(labels)
+        if nr_lab > len(labels):
```
adelavega: You could raise this warning as early as `__init__`

pliers/extractors/audio.py
```diff
@@ -488,4 +494,140 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
     For details on argument specification visit:
     https://librosa.github.io/librosa/effect.html.'''
 
+
     _feature = 'percussive'
+
+
+YAMNET_INSTALL_MESSAGE = '''
```
adelavega: @tyarkoni any other thoughts on how to install?

this is not a python package, so pip can't be used. @rbroc talked about this, and we came up with adding this folder to the python path. You can also copy the contents to `site-packages`. Both work, but are at bit janky since it doesn't even have an `__init__.py` file.

An alternative is to provide a helper function is `support` but it seems like overkill and maybe a user problem given instructions are given?  That said, if a user choses to make it a package and install it properly, the way we import it here won't work, as it expects the `yamnet` directory to be part of the path (rather than a package *in* the path).

pliers/extractors/audio.py
rbroc: decided to pass only a `self.labels`, which will be = `all_labels` if `label_subset` is not set, otherwise it is `label_subset` minus all non-existing labels passed by the user.

In `_extract`, the final set of labels returned in the df is stored in an extra `labels` variable, which will be either the same values of `self.labels` sorted by average probability across time points, or a subset of that (still sorted) if `top_n` is passed.

pliers/extractors/audio.py
rbroc: `__init__` now raises an error if both `label_subset` and `top_n` are not `None`.
In `__init__`, a `self.label_idx` variable is also computed which has the indices of the labels in `label_subset` (if this exists), else it has all label indices.
This is used to subset the set of predictions and labels in `_extract` without using conditional statements checking whether `label_subset` argument has been passed.

If `top_n` is defined, outputs and labels are further subsetted in `_extract`.

Note that no matter whether `top_n` is defined or not, outputs are always returned so to have labels with higher predictions on avg appear earlier (left in wide format; lines 619-622).

rbroc:
@adelavega thanks for the 🚀 review! The latest commits should address all your suggestions.

pliers/extractors/__init__.py
```diff
@@ -52,7 +52,8 @@
                     TempoExtractor,
                     BeatTrackExtractor,
                     HarmonicExtractor,
-                    PercussiveExtractor)
+                    PercussiveExtractor,
+                    AudiosetLabelExtractor)
```
adelavega: You need to add this to `__all__` below

pliers/support/exceptions.py
```diff
@@ -36,6 +36,9 @@ class MissingDependencyError(PliersError):
     """
 
     def __init__(self, dependencies, message=MISSING_DEPENDENCY_MESSAGE,
-                 *args, **kwargs):
-        msg = message % ', '.join(dependencies)
+                 custom_message=None, *args, **kwargs):
+        if custom_message:
```
tyarkoni: You probably already know this, and I don't actually have any objection to the more explicit syntax, but you could use a ternary conditional here: 
```python
msg = custom_message if custom_message else message % ', '.join(dependencies)
```

Actually, in this particular case, you can even write:
```python
msg = custom_message or message % ', '.join(dependencies)
```

adelavega:
This is not a show stopper but I just got the following error on `SchematicNarrative`. Didn't happen for any other datasets, so I'm not sure why.

After extracting, I merge results from various stimuli. Usually from various extractors. 
FWIW, this is a backup method we implemented long ago and I never use.

Will look into it in more detail soon but just FYI

```                                                                                                                                                                                            
/neuroscout/neuroscout/populate/extract.py in _to_csv(results, dataset_name, task_name)                                                                                                                      
     71     """ Save extracted Pliers results to file. """                                                                                                                                                   
     72     if results != [] and 'EXTRACTION_DIR' in current_app.config:                                                                                                                                     
---> 73         results_df = merge_results(list(zip(*results))[1])                                                                                                                                           
     74         outfile = Path(                                                                                                                                                                              
     75             current_app.config['EXTRACTION_DIR']) / '{}_{}_{}.csv'.format(                                                                                                                           
                                                                                                                                                                                                             
/neuroscout/src/pliers/pliers/extractors/base.py in merge_results(results, format, timing, metadata, extractor_names, object_id, aggfunc, invalid_results, **to_df_kwargs)                                   
    277             dfs.append(r.to_df(timing=_timing, metadata=metadata,                                                                                                                                    
    278                                format='long', extractor_name=True,                                                                                                                                   
--> 279                                object_id=_object_id, **to_df_kwargs))                                                                                                                                
    280         elif invalid_results == 'fail':                                                                                                                                                              
    281             raise ValueError("At least one of the provided results was not an"                                                                                                                       
                                                                                                                                                                                                             
/neuroscout/src/pliers/pliers/extractors/base.py in to_df(self, timing, metadata, format, extractor_name, object_id, **to_df_kwargs)                                                                         
    163                               (np.isfinite(durations).any() or                                                                                                                                       
    164                                np.isfinite(orders).any())):                                                                                                                                          
--> 165             df.insert(0, 'onset', onsets)                                                                                                                                                            
    166             df.insert(0, 'duration', durations)                                                                                                                                                      
    167             df.insert(0, 'order', orders)                                                                                                                                                            
                                                                                                                                                                                                             
/usr/local/lib/python3.6/site-packages/pandas/core/frame.py in insert(self, loc, column, value, allow_duplicates)                                                                                            
   3215         """                                                                                                                                                                                          
   3216         self._ensure_valid_index(value)                                                                                                                                                              
-> 3217         value = self._sanitize_column(column, value, broadcast=False)                                                                                                                                
   3218         self._data.insert(loc, column, value,                                                                                                                                                        
   3219                           allow_duplicates=allow_duplicates)                                                                                                                                         
                                                                                                                                                                                                             
/usr/local/lib/python3.6/site-packages/pandas/core/frame.py in _sanitize_column(self, key, value, broadcast)                                                                                                 
   3386                                                                                                                                                                                                      
   3387             # turn me into an ndarray                                                                                                                                                                
-> 3388             value = _sanitize_index(value, self.index, copy=False)                                                                                                                                   
   3389             if not isinstance(value, (np.ndarray, Index)):                                                                                                                                           
   3390                 if isinstance(value, list) and len(value) > 0:                                                                                                                                       
                                                                                                                                                                                                             
/usr/local/lib/python3.6/site-packages/pandas/core/series.py in _sanitize_index(data, index, copy)                                                                                                           
   3996                                                                                                                                                                                                      
   3997     if len(data) != len(index):                                                                                                                                                                      
-> 3998         raise ValueError('Length of values does not match length of ' 'index')                                                                                                                       
   3999                                                                                                                                                                                                      
   4000     if isinstance(data, ABCIndexClass) and not copy:                                                                                                                                                 
                                                                                                                                                                                                             
ValueError: Length of values does not match length of index  
```

adelavega:
It's actually not `merge_results`. It's an error when calling `result.data`, because onsets and shape of data does not match up:

```
In [37]: r.onset.shape
Out[37]: (1852,)

In [38]: r._data.shape
Out[38]: (1851, 521)
```

adelavega:
It's failing on this file fyi: `'/file-data/stimuli/26897ea43186d06de746cbc8490024f9a617a955.wav'`

adelavega:
We determined this bug is due to the fact that one more onset is being included than should be. This does not occur when creating the data, thus the shorter length. 

A hacky workaround is to index the onsets by the length of the data: 

```
        onsets = onsets[:, preds.shape[0]]
```

rbroc:
> We determined this bug is due to the fact that one more onset is being included than should be. This does not occur when creating the data, thus the shorter length.
> 
> A hacky workaround is to index the onsets by the length of the data:
> 
> ```
>         onsets = onsets[:, preds.shape[0]]
> ```

there might be a more general timing issue. 
Will look into that asap! :)

pliers/extractors/audio.py
```diff
@@ -488,4 +494,101 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+
+        MODULE_PATH = path.dirname(yamnet.__file__)
+        WEIGHTS_PATH = path.join(MODULE_PATH, 'yamnet.h5')
+        LABELS_PATH = path.join(MODULE_PATH, 'yamnet_class_map.csv')
+
+        self.hop_size = hop_size
+        self.yamnet_kwargs = yamnet_kwargs
+        self.params = yamnet.params
+        self.params.PATCH_HOP_SECONDS = hop_size
+        for par, v in self.yamnet_kwargs.items():
```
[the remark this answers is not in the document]
rbroc: The yamnet model expects params to be a module with attributes, rather than a dictionary. 
Leaning towards leaving it as it is instead of reconverting back to this format within _extract. Simplified the logic a bit by removing the conditional statement (adding extra stuff to params shouldn't matter, it would simply not be used).

rbroc:
> We determined this bug is due to the fact that one more onset is being included than should be. This does not occur when creating the data, thus the shorter length.
> 
> A hacky workaround is to index the onsets by the length of the data:
> 
> ```
>         onsets = onsets[:, preds.shape[0]]
> ```

Onsets/duration issue should be fixed.

pliers/extractors/audio.py
rbroc: added `setup_yamnet` to support which should do the job. Also makes installation on travis smoother.

pliers/extractors/audio.py
```diff
@@ -488,4 +495,126 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+    hop_size (float): size of the audio segment (in seconds) on which label 
+        extraction is performed.
+    top_n (int): specifies how many of the highest label probabilities are 
+        returned. If None, all labels (or all in labels) are returned.
+        Top_n and labels are mutually exclusive arguments.
+    labels (list): specifies subset of labels for which probabilities 
+        are to be returned. If None, all labels (or top_n) are returned.
+        The full list of labels is available in the audioset/yamnet 
+        repository (see yamnet_class_map.csv).
+    weights_path (optional): full path to model weights file. If not provided,
```
tyarkoni: The `yamnet_path` argument is missing from the docstring.

pliers/extractors/audio.py
```diff
@@ -488,4 +495,126 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+    yamnet_kwargs (optional): Optional named arguments that modify input 
+        parameters for the model (see params.py file in yamnet repository)
+    '''
+
+    _log_attributes = ('hop_size', 'top_n', 'labels', 'weights_path',
+                       'yamnet_kwargs')
+
+    def __init__(self, hop_size=0.1, top_n=None, labels=None,
+                 weights_path=None, yamnet_path=None, **yamnet_kwargs):
+        if yamnet_path is None:
```
tyarkoni: Is there a reason to allow the user to specify a custom yamnet path? I'm assuming this would be an alternate way to import yamnet... the risk there is that users probably won't want to use this unless they have reason to think something has changed in the yamnet implementation. And if something *has* changed in the yamnet implementation, there's a good chance it will imply an API change, which could result in things breaking for us. So I guess my inclination would be to remove this check and always point to our own installer.

pliers/extractors/audio.py
```diff
@@ -488,4 +495,126 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+            yamnet_path = YAMNET_PATH
+        try:
+            sys.path.insert(0, str(yamnet_path))
+            self.yamnet = attempt_to_import('yamnet')
+            verify_dependencies(['yamnet'])
+        except MissingDependencyError:
```
tyarkoni: Why not just automatically install it for the user if it's missing?

pliers/extractors/audio.py
```diff
@@ -488,4 +495,126 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+        try:
+            sys.path.insert(0, str(yamnet_path))
+            self.yamnet = attempt_to_import('yamnet')
+            verify_dependencies(['yamnet'])
+        except MissingDependencyError: 
+            msg = ('Yamnet could not be imported. To download and set up '
+                  'yamnet, run:\n\tpython -m pliers.support.setup_yamnet')
+            raise MissingDependencyError(dependencies=None,
+                                         custom_message=msg)
+        verify_dependencies(['tensorflow'])
```
tyarkoni: I'd probably move the check for tensorflow ahead of the yamnet check, since the former is a prerequisite.

pliers/extractors/audio.py
```diff
@@ -488,4 +495,126 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+            missing = list(set(labels) - set(all_labels))
+            labels = list(set(labels) & set(all_labels))
+            if missing:
+                logging.warning(f'Labels {missing} do not exist. Dropping.')
+            self.labels = labels
+            self.label_idx = [i for i, l in enumerate(all_labels) 
+                              if l in labels]
+        else:
+            self.labels = all_labels
+            self.label_idx = range(len(all_labels))
```
tyarkoni: `range()` will return a generator, so depending on what you're doing with this later (or just to play it safe), you might want to convert it to a list at this point. If it's just passed through to `ExtractorResult` then it's probably fine.

pliers/extractors/audio.py
```diff
@@ -488,4 +495,126 @@ class PercussiveExtractor(LibrosaFeatureExtractor):
...
+                'The sampling rate of your stimulus '
+                f'({self.params.SAMPLE_RATE}Hz) must be at least twice the '
+                f'value of MEL_MAX_HZ ({self.params.MEL_MAX_HZ}Hz). Upsample'
+                ' your audio stimulus (recommended) or pass a lower value of '
+                'MEL_MAX_HZ when initializing the extractor.')
+
+        model = self.yamnet.yamnet_frames_model(self.params)
+        model.load_weights(self.weights_path)
+        preds, _ = model.predict_on_batch(np.reshape(stim.data, [1,-1]))
+        preds = preds.numpy()[:,self.label_idx]
```
tyarkoni: Not sure if numpy indexing works with generators; see comment above re: `label_idx`.

pliers/support/setup_yamnet.py
```diff
@@ -0,0 +1,41 @@
+import os
+from io import BytesIO
+from zipfile import ZipFile
+from urllib import request
+from pathlib import Path
+import sys
+import runpy
+
+DOWNLOAD_PATH = Path.home() / 'pliers_data'
+YAMNET_PATH = DOWNLOAD_PATH / 'models-master' / 'research' / 'audioset' / 'yamnet'
```
tyarkoni: I assume this mirrors the structure in the yamnet repo, but it's a lot of nesting. Unless the yamnet code requires that structure, I'd probably just store this under `.pliers/models/yamnet` or something like that (note that I believe we already use the convention of storing data in `.pliers` in other places, which is what many other packages do as well).

pliers/tests/extractors/test_audio_extractors.py
```diff
@@ -352,3 +357,60 @@ def test_percussion_extractor():
     assert np.isclose(df['onset'][29], 1.346757)
     assert np.isclose(df['duration'][29], 0.04644)
     assert np.isclose(df['percussive'][29], 0.004497, rtol=1e-4)
+
+
+@pytest.mark.parametrize('hop_size', [0.1, 1])
+@pytest.mark.parametrize('top_n', [5, 10])
+@pytest.mark.parametrize('target_sr', [22000, 14000])
+def test_audioset_extractor(hop_size, top_n, target_sr, yamnet_path=None):
```
tyarkoni: Does this currently test for alternate values of `yamnet_path`? If not, that seems worth adding.

pliers/extractors/audio.py
adelavega: Agreed, unless we think sometimes our install would fail. But I think that should be considered a bug to fix, so no need to build a workaround yet.

pliers/extractors/audio.py
adelavega: I think that's a little heavy handed and inconsistent with how pliers handles other `MissingDependency` errors, such as the extra downloads needed for librosa.

pliers/support/setup_yamnet.py
adelavega: I think it's that the whole TF repo is being downloaded. That said, in the download script you could download a subset, or download to a tmp dir, and only copy over what you need to `.pliers`

pliers/extractors/audio.py
rbroc: converted to list here.

pliers/extractors/audio.py
rbroc: turned `label_idx` into list (although indexing with generator was no problem)

pliers/extractors/audio.py
rbroc: > I think that's a little heavy handed and inconsistent with how pliers handles other `MissingDependency` errors, such as the extra downloads needed for librosa.
Yep, throwing a missing dependency error is consistent with what we do also, e.g., for nltk. I'd  lean towards letting the user trigger the download (I guess that is the point of having a `MissingDependencyError` in general), since it's also 125Mb of stuff. But no big deal to edit this so to do it for them, if you have a strong take on it.

pliers/extractors/audio.py
rbroc: removed the option to specify custom `yamnet_path` (that was only to avoid downloading extra stuff in the user already has the yamnet or the whole tf repo somewhere else).

pliers/tests/extractors/test_audio_extractors.py
rbroc: Having removed the yament_path argument, I guess this is now outdated.

pliers/support/setup_yamnet.py
rbroc: It is now so that the download script fetches the TF repo's zip archive (no way, as far as I can see, to only download a subset), extracts it all into a `yamnet_temp` folder in `pliers_data`, moves the yamnet part to a flat directory `pliers_data/yamnet` and deletes `yamnet_temp`.
We could in principle avoid extracting the whole thing, and just looping through files in the zip archive so to only extract the yamnet ones. That would however add a bunch of lines of code, as we would: 1) have to loop; 2) take all the extracted files and move them to a flat folder, as zipfile keeps the file structure when using `extract`. And after all, the whole thing is only 125Mb.
Just to make sure though, you are talking about `pliers_data` (same default folder in user's home directory as text dictionaries are dumped to), right? Not `.pliers`...

rbroc:
@tyarkoni @adelavega thanks a lot for your comments! The last few commits should address all of them, you're welcome to take a look and make sure it is so. 
I've made some slightly more substantial changes to `setup_yamnet.py` script, so to only keep the relevant folder and delete all the other TF models (see my reply to Tal's comment).

pliers/support/setup_yamnet.py
tyarkoni: Oh, I was under the impression we're putting stuff in `.pliers`, but if we're currently putting it in `pliers_data`, then yeah, putting this there as well makes sense.

pliers/extractors/audio.py
tyarkoni: Probably what we should ultimately do is settle on a unified way of handling downloaded models/files, so that users can do something like `pliers install yamnet --dir=/path/to/put/files` from the command line, where the storage location is optional (and would default to what we're currently using). Might be worth opening an issue. But since this is the only model like this right now (all the others have their own packages that handle setup themselves), let's not get sidetracked with that right now. This seems fine.

rbroc:
> LGTM! Okay to merge?

Should be, unless @adelavega has further comments.

adelavega:
Nope, LGTM. Merging!

PR #388, opened 2020-03-25
Fix index cols for no _to_df

adelavega:
A small bug in the `to_df `function which was causing incorrect `df` for some Extractors (`Clarifai` in this instance).

Only applies to those Extractors with no `_to_df` function and outputs in long format. 

I accidentally made all the columns index_cols include the feature names itself for that case.

### DIFF
--- a/pliers/extractors/base.py
+++ b/pliers/extractors/base.py
@@ -125,7 +125,7 @@ class ExtractorResult(object):
         if features is not None:
             index_cols = list(set(df.columns) - set(features))
         else:
-            index_cols = list(set(df.columns))
+            index_cols = []
 
         if hasattr(self, '_onsets'):
             onsets = np.array(self._onsets)


adelavega:
By the way, I see that pliers is also affected by a recent change very annoying change to travis. 

It seems that all "legacy" repos (which are not using Github Apps) have stopped working on travis (at least neuroscout and pyNS).

I can't figure out how to turn it back on though...

adelavega:
Looks like its this issue: https://www.traviscistatus.com/incidents/rx6fhs3wqcln

Builds are still running but status is not being updated on PRs

adelavega:
Merging

adelavega:
@tyarkoni if you have no objections i'm going to merge tonight

tyarkoni:
LGTM, thanks!

PR #389, opened 2020-04-02
Option to return log attributes in to_df

rbroc:
Adding the option to return `log_attributes` when calling `to_df` or `merge_results`.
`Log_attributes` are retrieved from `ExtractorResult.history.transformer_params`.


rbroc:
@tyarkoni at the moment this returns `log_attributes` in one column as serialized dictionary, rather than one column per attribute.
Not sure what the ideal way to return them would be when using merge_results with 'wide' format. Right now, if `log_attributes=True`, `merge_results` returns a separate `log_attribute` column per extractor and feature, with columns named `ExtractorName#FeatureName#log_attributes`. Open to suggestions for more elegant ways of doing this, e.g. one `log_attribute` column only with a dictionary having Extractor-Feature combinations as keys and `log_attributes` as values?

adelavega:
Can you check the format=`long` results when you get a chance?

adelavega:
Well there's tests for it, so should be OK, but manually check if you get a chance please! :)

rbroc:
> Well there's tests for it, so should be OK, but manually check if you get a chance please! :)

mh... anything specific you're thinking about? Seems to be working to me, but might be missing something.

tyarkoni:
Let's call the `to__df()` argument for this `estimator_params` rather than `log_attributes`, I think the former is clearer. Returning a single dict seems fine to me. And I think your proposal for the behavior in wide format is also reasonable; I don't think we want to make the user have to start parsing keys in a dict for different extractors; this is a bit messy, but should be pretty easily understood by users. Thanks!

rbroc:
renamed argument and columns to `estimator_params`.

adelavega:
Nothing in particular just checking. LGTM

tyarkoni:
Ugh, sorry, I meant to suggest renaming to `extractor_params`, not to `estimator_params` (I think I was working on pymare stuff around the time I reviewed that, and had estimators on the brain). Once you update that, LGTM!

rbroc:
> Ugh, sorry, I meant to suggest renaming to `extractor_params`, not to `estimator_params` (I think I was working on pymare stuff around the time I reviewed that, and had estimators on the brain). Once you update that, LGTM!

@tyarkoni edited that, but now the tests mysteriously fail at `pip install` stage.

tyarkoni:
Probably  because I delete the travis cache... maybe there's some change to the travis config that broke things because we were previously relying on cached packages?

adelavega:
Actually, that's a problem with `duecredit` (came up in fitlins) and should be fixed now. Restarting the tests...

adelavega:
Actually, that's a problem with `duecredit` (came up in fitlins) and should be fixed now. 

For some reason, travis is not letting me restart the tests, but that should fix it.

rbroc:
tried restarting tests by pushing a commit but not working.

rbroc:
Now magically restarted and tests are passing.

adelavega:
Great, i'll go ahead and merge then.

PR #390, opened 2020-04-20
add MetricExtractor

rbroc:
Add extractor taking probability distribution as input and returning metrics of the distribution (e.g. entropy)


rbroc:
@tyarkoni @adelavega  I’ve started looking into the entropy/other-probability-metric extractor in this PR. First step is creating a new stimulus type, that would be the input for the `ProbDistMetricExtractor`. We talked about something like a `ProbStim` or a `DictStim`, but I’m not sure where either of the two would fit in the `Stim` taxonomy, since stimuli are organized by modality.

The way I was thinking about it would be to create a new type of stimuli which are more "numeric/algebraic" in type, something like `VectorStims` (or `MatrixStim`) where the data is numpy vectors/arrays and an optional label attribute stores labels for each vector (if that applies). Extractors like the `ProbDistMetricExtractor` (will come up with a better name) could apply to these types, but other stuff like extractors computing autoregressive features of time series could be another example. Probability distributions could either simply be a `VectorStim` or have a specific `ProbDistStim` subclass. 

Open to suggestions on whether that could work, or whether you'd rather do things differently.

rbroc:
@adelavega @tyarkoni This should now be ready for a first review. I've implemented new stimulus and extractor (the way I explained in previous comment), and added tests for both which pass locally.
Can't see my latest commits here, something weird must have happened while Github was down. But all commits are there in the branch of my forked repo this PR comes from.

pliers/stimuli/vector.py
```diff
@@ -0,0 +1,109 @@
...
+    ''' Vector stimulus (1-D numpy array)
+
+    Args:
+         array (np.ndarray, list or pd.Series): Vector of values. Can be list, 
+            array or pandas Series. Gets converted to 1-D numpy array.
+        labels (list): List of labels to which probability values refer, if 
+            that applies.
+        data_dict (dict): If defined, overwrites array and labels. Keys are 
+            passed to labels argument, values are passed to array argument.
+        filename (str): Path to file, if array has to be read from file.
```
adelavega, line 19: I wonder if this stim should have so many options for loading, or if should keep it simple and leave that to the user (they can load, and then create stim)

pliers/stimuli/vector.py
```diff
@@ -0,0 +1,109 @@
...
+        if sort_data == 'descending':
+            labels = labels[::-1]
+            array = array[::-1]
+        return array, labels
+
+    @property
+    def data(self):
+        return self.array
+
+    def save(self, path, as_json=True):
```
adelavega, line 96: I'd think tsv might be a better default?

pliers/stimuli/vector.py
```diff
@@ -0,0 +1,109 @@
...
+    '''
+
+    _default_file_extension='.json'
+
+    def __init__(self, array=None, labels=None, filename=None, data_dict=None,
+        data_column=None, label_column=None, onset=None, duration=None, 
+        order=None, name=None, sort_data=None, url=None):
+
+        if filename is not None:
+            if filename.endswith('.json'):
```
adelavega, line 48: Sometimes filename endings are not sufficient to distinguish file type. Probably edge case though.

pliers/stimuli/vector.py
```diff
@@ -0,0 +1,109 @@
...
+                with open(filename) as f:
+                    data_dict = json.loads(f.read())
+            else:
+                df = pd.read_csv(filename, sep='\t')
```
adelavega, line 52: I guess one reason I would prefer to keep this simple and let the user do the reading, is that you might want to load in a different way, and it's a hassle to support all the relevant kwargs. 

For example, reading a csv with commas as separators.

pliers/extractors/vector.py
```diff
@@ -0,0 +1,67 @@
...
+    def __init__(self, functions=['np.mean', 'np.std', 'np.max', 'np.min'], 
+                 var_names=None, **kwargs):
+        functions = listify(functions)
+        if var_names is not None:
+            if len(var_names) != len(functions):
+                raise ValueError('Length or var_names must match number of '
+                                 'functions')
+        for idx, f in enumerate(functions):
+            if type(f) == str:
+                f = f.replace('numpy', 'np')
```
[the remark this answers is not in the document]
rbroc: It's simply because numpy is already imported as np here, so forcing to to np would avoid having to reimport with full package name specification.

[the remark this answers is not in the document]
adelavega: Yes, that made sense to me. My point was that more generally `np` is not an "official" shortname for numpy, just a convention. So it might be safer and more genral to just generalize this to any package the user may have installed, and require the complete package name, etc..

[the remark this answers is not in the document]
rbroc: okay, so that implies we would try to import whatever package the function comes from? Makes sense, if it is not a problem to potentially reimport numpy here without shortname.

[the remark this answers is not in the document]
rbroc: read now you comment below, which makes things clear.

pliers/stimuli/vector.py
rbroc: leaving read from and save to tsv option only (reads both from `filename` and `url`).

pliers/stimuli/vector.py
rbroc: forcing to read from tsv.

pliers/extractors/vector.py
```diff
@@ -0,0 +1,67 @@
...
+            eval() method).
+        var_names (optional): list of custom alias names for each metric
+        kwargs: named arguments for function call
+    ''' 
+
+    _input_type = VectorStim
+    _log_attributes = ('functions',)
+
+    def __init__(self, functions=['np.mean', 'np.std', 'np.max', 'np.min'], 
+                 var_names=None, **kwargs):
```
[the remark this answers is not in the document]
rbroc: I guess the alternative would be asking the user to pass a dictionary of dictionaries with function names as keys and kwargs dictionary as values. 
Leaving it as it is for now, but open to alternative options :)

pliers/extractors/vector.py
```diff
@@ -0,0 +1,67 @@
...
+        if var_names is not None:
+            if len(var_names) != len(functions):
+                raise ValueError('Length or var_names must match number of '
+                                 'functions')
+        for idx, f in enumerate(functions):
+            if type(f) == str:
+                f = f.replace('numpy', 'np')
+                try:
+                    f_list = f.split('.')
+                    if len(f_list) > 1 and 'lambda' not in f:
```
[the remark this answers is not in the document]
rbroc: now importing via `import_module`. Also removed option to evaluate lambda (user can define function then pass it directly, I guess).

rbroc:
done implementing @adelavega's suggestions (thank you!) and ready for review again. Tests seem to pass if you look at the travis log, but coverall times out for some weird reason.

rbroc:
@tyarkoni @adelavega I've rebased this to [Tal's PR](https://github.com/tyarkoni/pliers/pull/394) and:
- Removed `VectorStim` and related tests
- Moved `MetricExtractor` to new `misc` type and set `SeriesStim` as input
- Added option to subset `SeriesStim.data` in input to `MetricExtractor` by passing list of Series indices as argument (added test for this)
- Fixed a little bug in `BertLM`. Tokens are capitalized before being used as column names (to avoid overlap with `onset,` `duration` and other metadata columns). But this caused distinct one-character non-ascii tokens to be "capitalized" into the same character, and therefore collapsed into same column when calling `to_df. Now tokens containing non-ascii are left unchanged.
- Added test to make sure implicit conversion works when feeding `ExtractorResult` as input to `MetricExtractor`.
If we're not planning on creating a `DataFrameStim` within this PR, this is ready for review.

pliers/extractors/misc.py
```diff
@@ -0,0 +1,78 @@
...
+        functions
+    Args:
+        functions (str, functions or list): function or list of numpy, scipy, 
+            or custom functions to be applied to 1-dimensional numpy arrays.
+            Functions can be passed directly (e.g. passing scipy.stats.entropy) if 
+            the package/module they belong to has been imported, or as strings 
+            (e.g. 'numpy.mean'). Custom functions returning integers or iterables 
+            can also be passed by passing the function itself.
+        var_names (list): optional list of custom alias names for each metric
+        subset_idx (list): subset of Series indices to compute metric on.
```
adelavega, line 26: Could this be label based? On Tal's PR I commented that labels are not a requirement for pandas Series, but by default they do get added when using the converter. It seems quite clunky to have to create a boolean index and then pass it to this extractor, rather than labels. 

If we're going down this route, we may even want to allow for regex to filter columns. Although I think in most cases we'l want to restrict columns at the level of the first extractor's output.

pliers/extractors/misc.py
```diff
@@ -0,0 +1,78 @@
...
+        functions = listify(functions)
+        if var_names is not None:
+            var_names = listify(var_names)
+            if len(var_names) != len(functions):
+                raise ValueError('Length or var_names must match number of '
+                                 'functions')
+        for idx, f in enumerate(functions):
+            if isinstance(f, str):
+                try:
+                    f_mod = f.split('.')[:-1]
```
adelavega, line 44:
```suggestion
                    f_mod, f_func = f.rsplit('.', 1) 
```

pliers/extractors/misc.py
```diff
@@ -0,0 +1,78 @@
...
+                    f_func = f.split('.')[-1]
+                    functions[idx] = getattr(import_module('.'.join(f_mod)),
```
adelavega, line 46:
```suggestion
                    functions[idx] = getattr(import_module(f_mod),
```

pliers/extractors/misc.py
rbroc: not sure I understand. The labels are added as index at conversion, right?

rbroc: right, this was just a terminology issue, should be solved by updated docstring.

rbroc:
implemented @adelavega's suggestions.

pliers/extractors/misc.py
```diff
@@ -0,0 +1,75 @@
...
+            if idx_diff:
+                logging.warning(f'{idx_diff} not in index.')
+            if not idx_int:
+                raise ValueError('No valid index')
+            series = stim.data[idx_int]
+        else:
+            series = stim.data
+        for f in self.functions:
+            metrics = f(series, **self.kwargs)
+            if isinstance(metrics, Iterable):
```
tyarkoni, line 70: One issue here is that strings are `Iterable`, so if a function returns a string, it'll be turned into a list of characters. You can use `pliers.utils.isiterable()`, which will only check for certain types. I don't think it catches pandas series though, so we may want to add that to the type tuple in `isiterable`.

rbroc: added `pd.Series` to `isiterable` (noticed that it does not include numpy arrays, btw) and now using that to check for `type` of `metrics`.

rbroc:
implemented Tal's suggestions, @tyarkoni @adelavega this should be ready to merge.

Issue #391, opened 2020-04-27
Drop vestigial Python 2 features


tyarkoni:
We already officially only support and test Python 3, but we still have vestigial Py2 code all over the place (most notably, use of `six`). It's probably worth doing a sweep through the codebase and removing as much as possible of the Py2/3 compatibility code to reduce the maintenance burden.

See also: #323

adelavega:
Probably worth also uniformly using pathlib as well

hugovk:
Oops, looks like #393 overlaps with #392!

tyarkoni:
@adelavega agreed, but that can be a separate PR.

tyarkoni:
@hugovk I think we need to decide whether or not we're dropping 3.5 support. I incline not to unless we're already using 3.6+ features, which I don't think is the case.

It looks like the main difference between PRs, aside from programmatic changes via `pyupgrade`, is that I re-ordered import statements everywhere. Unless you have objections, I think I'll merge #392, and if we decide to limit to 3.6+, you can open a new PR for that. Sound reasonable?

hugovk:
Yes, that's reasonable.

Dropping 3.5 is a separate conversation, distinct from this issue.

tyarkoni:
Okay, I merged #392. You're welcome to open a PR with other `pyupgrade` changes I didn't make manually (e.g., updating `super()` calls) as long as they're 3.5 compatible. Thanks for the PR, and sorry about the confusion—I should have immediately self-assigned when I opened the issue.

hugovk:
Thanks, I've rebased it into https://github.com/tyarkoni/pliers/pull/393 and dropped the commit to upgrade to 3.6+ syntax.

tyarkoni:
Great, thanks! I'll merge as soon as the tests pass.

PR #393, opened 2020-04-27
Upgrade Python syntax and add python_requires for 3.5+

hugovk:
Fixes https://github.com/tyarkoni/pliers/issues/391.

* Uses https://github.com/asottile/pyupgrade to upgrade Python syntax for Python 3.5+
* Add `python_requires='>=3.5'` metadata to help pip

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,5 +1,4 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
 #
 # pliers documentation build configuration file, created by
 # sphinx-quickstart on Wed Dec 27 16:15:14 2017.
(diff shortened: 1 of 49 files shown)


tyarkoni:
Eep, I just opened #392...

Aside from minor differences between PRs, I'm not sure we want to drop 3.5 support yet given that ~20% of people are still on it. Do people have thoughts?

hugovk:
Ah! I thought Python 3.5 had already been dropped in https://github.com/tyarkoni/pliers/pull/372, but that must just be the testing of 3.5.

Python 3.5 is still supported by core Python team until later this year, so it's fine to still keep it.

Although, NEP 29 is something to consider: https://numpy.org/neps/nep-0029-deprecation_policy.html

Let me know, and I'll drop that change!

tyarkoni:
Thanks!

PR #397, opened 2020-05-22
fix bertLM onset/duration

rbroc:
fixing indexing issue with BertLMExtractor onset/duration.

### DIFF
--- a/pliers/converters/misc.py
+++ b/pliers/converters/misc.py
@@ -13,6 +13,8 @@ class ExtractorResultToSeriesConverter(Converter):
 
     def _convert(self, result):
         df = result.to_df(timing=False, metadata=False, object_id=False)
+        if 'object_id' in df.columns:
+            df = df.drop(['object_id'], axis=1)
         stims = []
         for i, data in df.iterrows():
             onset = result.onset[i] if result.onset is not None else None
--- a/pliers/extractors/audio.py
+++ b/pliers/extractors/audio.py
@@ -602,7 +602,7 @@ class AudiosetLabelExtractor(AudioExtractor):
         model = self.yamnet.yamnet_frames_model(self.params)
         model.load_weights(self.weights_path)
         preds, _ = model.predict_on_batch(np.reshape(stim.data, [1,-1]))
-        preds = preds.numpy()[:,self.label_idx]
+        preds = preds[:,self.label_idx]
         
         nr_lab = self.top_n or len(self.labels)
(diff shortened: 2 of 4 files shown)


rbroc:
it seems like this is fixing the problem with BertLM onsets. 
@adelavega, if you confirm that it is the case, this should be ready to merge.

tyarkoni:
@adelavega I'm slammed with grants and reviews for the next 2 weeks, so feel free to merge pliers PRs as you see fit.

adelavega:
i'll give this a closer look tomorrow, but it's looking good.

rbroc:
ok, I've noticed there was another little bug, namely that `object_id` should not be included in the `data` field of the resulting `stim` when using `ExtractorResultToSeriesStim`. Otherwise, the value of `object_id` will be included in the metric computation when `SeriesStim` is passed to `MetricExtractor`.

This was already the case for `ExtractorResults` from Extractors that do not have a custom `_to_df` methods, but we needed an additional bit of logic to generalize to extractors with custom `_to_df`.

pliers/converters/misc.py
```diff
@@ -13,6 +13,8 @@ class ExtractorResultToSeriesConverter(Converter):
 
     def _convert(self, result):
         df = result.to_df(timing=False, metadata=False, object_id=False)
```
adelavega, line 15: I still don't get why `object_id=False` would not guarantee that `object_id` is not in the columns. Is it because `to_df` could be a custom function for the Extractor that does not respect that argument?

rbroc: It's because if a custom `_to_df` function is defined for an Extractor, `to_df` calls that first. If the custom function also adds an `object_id` column, there is no place in the code where that column is dropped in the `to_df` call. 
The object_id argument only makes it so that an object_id column is _added_ if not present.

Alternative solutions to fixing the problem this way could be to actually edit the to_df function so to drop `object_id` column if present when `object_id=False`.

Another option is changing the custom `_to_df` function for `BertLMExtractor`. 
At the moment, it inherits from `BertExtractor`, where adding a custom `object_id` is necessary. But  I guess we could define a `_to_df` for` BertLMExtractor` where `object_id` is not added.

Open to implement any of these if more elegant.

adelavega: I think the second option is a good idea but it's out of the scope of this PR so let's merge but and deal with it later.

PR #398, opened 2020-05-28
support lambda functions in MetricExtractor

rbroc:


rbroc:
This has a few limitations, e.g. that lambdas requiring modules other than numpy and scipy (i.e. those imported in the misc extractors module) cannot be evaluated.

adelavega:
My main concern here now that I've thought about it some more, is that it's a bit dangerous to allow arbitrary evaluation of strings. I'll look into it.

tyarkoni:
We're already allowing callables to be passed, so I don't think eval() introduces any new security risks. Anything malicious you could do in a string you could also do inside a callable. I think it's the user's responsibility not to run pliers in an insecure environment.

adelavega:
Allright lets do it then!

PR #399, opened 2020-06-08
fix labels indexing issue in Audioset extractor

rbroc:
Fixing a bug that caused mismatch between labels and values when passing subset of AudioSet labels to AudioSet extractor.

### DIFF
--- a/pliers/extractors/audio.py
+++ b/pliers/extractors/audio.py
@@ -10,6 +10,8 @@ from scipy import fft
 import pandas as pd
 import soundfile as sf
 
+import logging
+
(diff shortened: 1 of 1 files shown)


tyarkoni:
Hah, given the wonky results we've been seeing in NeuroScout (e.g., whistling), this is one bug I'm kind of happy to see! Hopefully it fixes the weirdness. :)

tyarkoni:
@rbroc LGTM assuming the tests pass.

adelavega:
Yep, the whistling result is the first thing that came to mind

Issue #403, opened 2020-07-06
Error using GoogleVisionAPIFaceExtractor


queenknight:
Hi! I'm new to coding and had an error trying out the GoogleVisionAPIFaceExtractor to identify the number of faces appearing each second on a 10min video. I had already set an environment variable for the API key.

**This is the code that I used:**

```
from pliers.stimuli import VideoStim
from pliers.filters import FrameSamplingFilter
from pliers.extractors import GoogleVisionAPIFaceExtractor, merge_results

video = VideoStim('/Users/queenknight/Documents/movie.mp4')
sampler = FrameSamplingFilter(hertz=1)
frames = sampler.transform(video)

ext = GoogleVisionAPIFaceExtractor()
results = ext.transform(frames)
df = merge_results(results, )

df.to_excel('moviegoogleface.xlsx', index=False, header=True)
```

**This was the error I got:**
```
638it [47:07,  4.43s/it]
Stim: 638it [00:00, 73165.80it/s]
Traceback (most recent call last):
  File "/Users/queenknight/PycharmProjects/Movie/Faceextractor_movie.py", line 13, in <module>
    df = merge_results(results, )
  File "/Users/queenknight/OneDrive/EDX/lib/python3.8/site-packages/pliers/extractors/base.py", line 271, in merge_results
    dfs.append(r.to_df(timing=_timing, metadata=metadata,
  File "/Users/queenknight/OneDrive/EDX/lib/python3.8/site-packages/pliers/extractors/base.py", line 114, in to_df
    df = self.extractor._to_df(self, **to_df_kwargs)
  File "/Users/queenknight/OneDrive/EDX/lib/python3.8/site-packages/pliers/extractors/api/google.py", line 80, in _to_df
    name = 'landmark_' + lm['type'] + '_%s'
KeyError: 'type'
```

**May I know if this is a bug or an error in my code? Would appreciate any solution to fix it, thank you!**
Unfortunately, I see that the Google Cloud Vision API recorded the number of my requests despite the error, eating into the free trial.

tyarkoni:
@queenknight thanks for the report. I haven't had a chance to look into this yet, but my initial guess based on the error is that the Google face APIs may have changed the structure of the results it returns, potentially breaking our extractor classes. 

[...]

PR #406, opened 2020-07-15
FIX: If Google Face Extractor result landmark has no type, skip. 

adelavega:
This should fix the issue with Google Face extractor. It seems sometimes it returned landmark annotations without a type. I just chose to skip them.


adelavega:
Closes #403

PR #410, opened 2020-07-16
Remove instance of Python2 super

### LINKED ISSUE #371: Autoreload does not work with pliers
IPython `%autoreload` leads to a strange error.

I think it might be related to: https://stackoverflow.com/questions/32481508/autoreload-and-package-causing-typeerror-supertype-obj-obj-must-be-an-insta

The second answer seems to suggest that changing all of the `super` calls in Pliers to Python 3 format may fix this problem.

`super().__init__()` instead of `super(Extractor, self).__init__() `

Not high priority, but it makes it impossible to use `autoreload` when developing.

adelavega:
Minor removal of vestigial Python 2 super call.

Fixes #371 

The rest was taken care of in a recent PR #392

### DIFF
--- a/pliers/extractors/misc.py
+++ b/pliers/extractors/misc.py
@@ -5,29 +5,28 @@ Extractors that operate on Miscellaneous Stims.
 from pliers.stimuli.misc import SeriesStim
 from pliers.extractors.base import Extractor, ExtractorResult
 from pliers.utils import listify, isiterable
-import scipy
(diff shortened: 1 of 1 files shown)


tyarkoni (approved):
Sorry, totally missed this. LGTM.

adelavega:
Oh, wow I had completely forgotten about this. LGT my current self as well.

Issue #347, opened 2019-05-09
docker image build failed locally


yarikoptic:
the check for being properly installed seems to fail
```
...
Step 7/8 : RUN python -m pliers.support.download
 ---> Running in 32a7381f92dd
/usr/local/bin/python: Error while finding module specification for 'pliers.support.download' (ModuleNotFoundError: No module named 'pliers.support')
The command '/bin/sh -c python -m pliers.support.download' returned a non-zero code: 1
$> git describe
0.2.3-174-g38e54d9
```

yarikoptic:
ha -- I think it could be the effect of my restrictive mask (outside):
```
pliers@cd5e50a7ea71:/opt/pliers$ ls -ld pliers/
drwx------ 1 root root 270 May  9 15:16 pliers/
```
I now vaguely remember running into a similar situation.  I will augment Dockerfile by hardcoding `umask 022` within the installation `RUN`

yarikoptic:
ah, `COPY . .` so it propagates my restrictive permissions.  Will augment by chmod after ;)  but also will cd away from that build directory when running the test

kaczmarj:
I wonder if —chown=root in the COPY command would fix this too...

Jakub

yarikoptic:
well, it is not about who owns (copy is owned by root) but rather who can read after you become another user (pliers) ;)  `chmod` worked for me, doing the final test build, will submit a PR

adelavega:
@yarikoptic any idea if this got closed with a PR or still need it open?

yarikoptic:
cannot test since now it fails with
```
$> docker build -t pliers-test-0.4.0-5-g3291840 .
...
Processing triggers for libc-bin (2.28-10) ...
Processing triggers for libgdk-pixbuf2.0-0:amd64 (2.38.1+dfsg-1) ...
ERROR: Could not open requirements file: [Errno 2] No such file or directory: 'requirements.txt'
[...]

PR #411, opened 2020-07-17
ENH: use multi-stage builds in Dockerfile to reduce final image size

kaczmarj:
This commit reduces the uncompressed image size by about 1 GB. A major change is that a python virtual environment is used within the Docker image. This is done so that all installed dependencies are below one directory, and this directory can be copied into the final build stage.

I built an image using this branch available at https://hub.docker.com/r/kaczmarj/pliers (`docker pull kaczmarj/pliers`).

I was inspired to update the dockerfile after the discussion in #347

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -1,4 +1,4 @@
-FROM python:3.6-slim as builder
+FROM python:3.7-slim as builder
 ARG DEBIAN_FRONTEND="noninteractive"
 WORKDIR /opt/pliers
 COPY . .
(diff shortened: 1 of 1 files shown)


adelavega:
Nice improvement. Did the tests pass inside of the docker image?

kaczmarj:
These tests passed

```
python -m pytest pliers/tests/test_* pliers/tests/converters pliers/tests/filters \
  --cov=pliers --cov-report= -m "not requires_payment" -W ignore::UserWarning
```

but these did not:

```
python -m pytest pliers/tests/extractors \
  --cov=pliers --cov-report= -m "not requires_payment" --cov-append -W ignore::UserWarning
```

Got 2 errors:

```
FAILED pliers/tests/extractors/test_misc_extractors.py::test_metric_extractor - TypeError: ufunc 'isfinite' not supported for the input types, and the inputs could not be sa...
FAILED pliers/tests/extractors/test_text_extractors.py::test_bert_extractor - AttributeError: module 'transformers' has no attribute 'TFAutoModel'
```

I don't get these two errors in the original container, so let me debug...

adelavega:
They both seem like versioning issue so hopefully they're easy to fix!

kaczmarj:
It turns out the savings in size were due to using a non-GPU build of tensorflow. The default `venv` has slightly older packages (`pip` and `setuptools`) by default, so for some dependency reason, tensorflow 1.x was installed instead of 2.x, and the 2.x wheels ship by default with gpu support. `tensorflow_cpu` can be installed instead to save a few hundred MB. i updated the packages in the venv.

In any case, I still think this multi-stage build is an improvement over the current Dockerfile, because it offers a clear separation between compilation and the final image. With BuildKit enabled (i.e., `DOCKER_BUILDKIT=1`), the two stages can be built in parallel. Not a massive increase in speed in the case the pliers docker image, but a nice feature nonetheless.

In the current state of this PR, all tests pass in the docker image:

```
docker run --rm -it --workdir /opt/pliers kaczmarj/pliers \
  bash -c 'pip install pytest-cov \
    && python -m pytest \
         pliers/tests/test_* pliers/tests/converters pliers/tests/filters pliers/tests/extractors \
         --cov=pliers --cov-report= -m "not requires_payment" --cov-append -W ignore::UserWarning'
```

adelavega:
Cool, thanks. Too bad about the space savings but yeah LGTM.

PR #414, opened 2020-09-11
[FIX] remove --no-user-group

### LINKED ISSUE #413: Dockerfile: chown group error
Docker version 19.03.12, build 48a66213fe

When building the Docker Image, I got the error (or something close to this error):
```
unable to convert uid/gid chown string to host mapping: can't find gid for group pliers: no such group
```

which occurs during [`COPY --from=builder --chown=pliers /opt/venv /opt/venv`](https://github.com/jdkent/pliers/blob/a541b07963f2d6eb45f12a52d3e6212171b03004/docker/Dockerfile#L45) because if a group is not specified in `--chown`, the version of docker I'm using assumes the group is the same as the user (e.g., `pliers`), but since [`RUN useradd --no-user-group --create-home --shell /bin/bash pliers`](https://github.com/jdkent/pliers/blob/a541b07963f2d6eb45f12a52d3e6212171b03004/docker/Dockerfile#L33) specifies that a group should not be added, then the error occurs since the `pliers` group does not exist.

I can fix it by removing the option `--no-user-group`, but I do not know what unintended side effects I'm incurring.

@kaczmarj: looks like you wrote those lines, would you know if removing `--no-user-group` is a bad idea?

jdkent:
closes #413

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -30,7 +30,7 @@ RUN python -m pip install --no-cache-dir ipython notebook
 RUN python -m spacy download en_core_web_sm
 
 FROM python:3.7-slim
-RUN useradd --no-user-group --create-home --shell /bin/bash pliers \
+RUN useradd --create-home --shell /bin/bash pliers \
     # Empty top level directories to facilitate use of the image in singularity
     # on a box with kernel lacking overlay FS support
     && mkdir -p /data /backup \


adelavega:
I don't know why that is in the original `Dockerfile` but I would say if you are able to get it to build, and get all the tests to pass within the container, it should be OK

jdkent:
Could you try rebuilding master? There are [a few failures](https://travis-ci.com/github/jdkent/pliers/jobs/384119534) that are not immediately obvious to me as to why this Dockerfile edit would make them fail.

adelavega:
I wonder why travis is not showing up here... it also doesn't seem to think i have those privileges

adelavega:
They're probably not related to the Dockerfile change, I'll open another PR with just a trivial change to trigger a build.

adelavega:
https://travis-ci.org/github/PsychoinformaticsLab/pliers

jdkent:
One thing that's presumably causing the ` AttributeError: module 'yamnet' has no attribute 'params'` error is [this line]( https://github.com/PsychoinformaticsLab/pliers/blob/5dedbe7dcd415b336942317eef020e69b84f926b/pliers/support/setup_yamnet.py#L14), I'll change the url to a release the represents the code when it was working.

adelavega:
that actually failed for me recently in neuroscout so that explains it

jdkent:
Now that the unrelated test failures have been addressed, does this pull request look okay?

adelavega:
LGTM

PR #418, opened 2020-09-12
[FIX] tests

### LINKED ISSUE #416: yamnet tests failing
related to #414  ([this comment](https://github.com/PsychoinformaticsLab/pliers/pull/414#issuecomment-691338476)).

Many tests that use yamnet have the error message:
```
AttributeError: module 'yamnet' has no attribute 'params'
```
most likely related to the version of tensorflow models being downloaded/installed in the pliers package.

### LINKED ISSUE #417: pytesseract tests failing
related to #414  ([this comment](https://github.com/PsychoinformaticsLab/pliers/pull/414#issuecomment-691338476)).

tests using pytesseract (0.3.5, 0.3.6) give different output than 0.3.4:

```
AssertionError: assert (0, 'text[Exit]') in array(['text[Exit\n\n]'], dtype=object)
```

Most notably the more recent versions appear to append newlines and other hidden characters to the text, pinning to version 0.3.4 appears to resolve the issue.

### LINKED ISSUE #419: a couple librosa tests failing
related to #414  ([this comment](https://github.com/PsychoinformaticsLab/pliers/pull/414#issuecomment-691338476)).

The outputs for `TonnetzExtractor` and `ChromaCQTExtractor` changed with librosa==0.8.0, resulting in a couple failed tests. I do not know what the ground truth _should_ be, so I'm falling back on reproducibility.

jdkent:
closes #416 
closes #417 
closes #419

This pull request makes versions for pytesseract, yamnet, and librosa more restrictive.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -48,8 +48,8 @@ before_script:
 - python -m pliers.support.download
 - python -m spacy download en_core_web_sm
 script:
-  - py.test pliers/tests/test_* pliers/tests/converters pliers/tests/filters --cov=pliers --cov-report= -m "not requires_payment" -W ignore::UserWarning
-  - py.test pliers/tests/extractors --cov=pliers --cov-report= -m "not requires_payment" --cov-append -W ignore::UserWarning
+  - travis_wait 30 py.test pliers/tests/test_* pliers/tests/converters pliers/tests/filters --cov=pliers --cov-report= -m "not requires_payment" -W ignore::UserWarning
+  - travis_wait 30 py.test pliers/tests/extractors --cov=pliers --cov-report= -m "not requires_payment" --cov-append -W ignore::UserWarning
 after_success:
 - coveralls
 before_cache:
--- a/optional-dependencies.txt
+++ b/optional-dependencies.txt
@@ -5,7 +5,7 @@ python-twitter
 gensim
 google-api-python-client
 google-compute-engine
-librosa>=0.6.3
+librosa>=0.6.3,<0.8.0
 numba<=0.48
 matplotlib
 opencv-python
--- a/optional-dependencies.txt
+++ b/optional-dependencies.txt
@@ -12,7 +12,7 @@ opencv-python
 pathos
 pygraphviz
 pysrt
-pytesseract
+pytesseract==0.3.4
 python-twitter
 scikit-learn
 seaborn
--- a/optional-dependencies.txt
+++ b/optional-dependencies.txt
@@ -5,14 +5,14 @@ python-twitter
 gensim
 google-api-python-client
 google-compute-engine
-librosa>=0.6.3,<0.8.0
+librosa>=0.6.3
 numba<=0.48
 matplotlib
 opencv-python
 pathos
 pygraphviz
 pysrt
-pytesseract==0.3.4
+pytesseract
 python-twitter
 scikit-learn
 seaborn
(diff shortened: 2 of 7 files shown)


tyarkoni:
Thanks for catching these! Haven't looked in detail at the issues, but if we can rewrite the tests to avoid pinning to old versions, that would be preferable. E.g., in the pytesseract case, we could (a) allow different versions to produce different outputs (and edit the test to only check that the beginning of the string matches the target), (b) edit the extractor to strip trailing whitespace, or (c) pin to the most recent version and require the whitespace.

My preference is probably to go with (a), even though it produces different output depending on dependency version, because it's a fairly benign difference, and there are probably principled reasons why the changes were introduced (e.g., I imagine that the idea was to preserve whitespace in the text in the original image).

jdkent:
Thanks for the suggestions! I changed the text searching tests to search for the "Exit" within the output instead of matching an exact string.

The librosa changes are a little more difficult since they are floating-point values and I do not know if increasing the error tolerance would be helpful, so I created a modified dictionary whose keys can be software ranges. That allowed me to state if librosa is between such and such version, I expect a specific floating-point number.
Using that strategy could allow us to track how changing dependency versions change outputs.

I'm keeping the tensorflow-models pinned 2.3.0 for now, but since they now have releases on pypi, the installation of the package could likely be streamlined..

rbroc:
> I'm keeping the tensorflow-models pinned 2.3.0 for now, but since they now have releases on pypi, the installation of the package could likely be streamlined.
> 

tensorflow research models (including yamnet) are sadly not yet part of tf-models nor of tf-hub, but let's definitely keep an eye out! :)

jdkent:
thanks for the clarification @rbroc! I went another route of only downloading the latest tagged release of the [tensorflow/models](https://github.com/tensorflow/models) repository instead of the master branch, since there was some sort of backward breaking change in their master branch. This way, pliers is not pinned to a specific version of the tensorflow/models repository while still relying on a tagged release (which should have some more guarantees for stability). 

Another downside of installing `tf-models-official` is all the extra dependencies it has pinned as well.

I think this pull request is over-engineered and ready for review.

tyarkoni:
LGTM but I don't want to merge until travis is being triggered. Do we know what's going on with that?

jdkent:
[It's building under my forked repository, but not this one](https://travis-ci.com/github/jdkent/pliers/builds/184466964)

adelavega:
It's building for this one too actually: https://travis-ci.org/github/PsychoinformaticsLab/pliers

It's just not showing up here. I bet it has to do w/ the move of domain from tyarkoni to here.

I would reset Travis by rejecting permissions, and then giving permissions again @tyarkoni 
I tried to but it wouldn't listen to me.

TBH, Travis has pulled this on me on and off over the last year and it's pretty annoying. They're trying to move everyone from travis.org to travis.com but it's not been a smooth transmission.

tyarkoni:
This might not be a bad impetus to migrate to GitHub Actions. I've heard good things from several people. Does anyone want to take a look and see what's involved? If not too much work, it might be worth it.

adelavega:
Yeah, I tried the same and nothing. I thought GitHub Actions was the default was of Travis to hookup now? It's a mess honestly.

tyarkoni:
Finally figured it out. It was a pretty convoluted process, but should be good now.

Re: GitHub Actions, they provide their own CI hooks (see, e.g., [this](https://developer.okta.com/blog/2020/05/18/travis-ci-to-github-actions)). Still think it might be worth exploring—if not now, then the next time we start a new project.

PR #421, opened 2020-09-18
update readme to link to correct pliers docs page

jsmentch:
The links to the docs 404 and I couldn't find them for a bit. The other links seem to forward to the correct places though


adelavega:
thanks!

PR #422, opened 2020-09-23
[FIX] access string in numpy array

jdkent:
looks like one of the changes in #418 [had a regression with the Google API](https://travis-ci.com/github/PsychoinformaticsLab/pliers/jobs/387471662#L1294),

This pull request aims to fix the regression.

### DIFF
--- a/pliers/tests/test_graph.py
+++ b/pliers/tests/test_graph.py
@@ -175,7 +175,7 @@ def test_small_pipeline_json_spec3():
     assert history.shape == (2, 8)
     assert history.iloc[0]['result_class'] == 'TextStim'
     result = merge_results(result, format='wide', extractor_names='multi')
-    assert 'Exit' in result['stim_name'].values[0]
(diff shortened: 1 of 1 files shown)


jdkent:
looks like my pull request does not have access to the same environmental variables as builds on master have resulting in some tests being skipped. Is this on purpose to ensure API call limits do not go over?

adelavega:
Yes, the API tests are not being run on PRs. 

I usually will test these locally once I think I'm close. You could do the same if you get some Google keys (there should be a free tier in the meantime).

jdkent:
Thanks for the clarification, [is this where I get some google keys](https://console.cloud.google.com/projectselector2/apis/credentials/serviceaccountkey)?

Always wanted to take a google on a test drive

adelavega:
yep! just get some there and put them in the environment

adelavega:
I tested this and it fixes the regression. Will merge, but I also noticed there's a lot more regressions w/ Google API that need to be looked into.

Issue #426, opened 2020-10-27
add downsampling filter


adelavega:
This should operate on extractor result outputs

tyarkoni:
This may take some work to properly integrate into the `Transformer` hierarchy... Currently every `Transformer` has to take a `Stim` as input, and `ExtractorResult` objects aren't Stims. We have the ER --> `SeriesStim` converter, so we could in principle convert every ER to a series, downsample via the new filter, and then re-aggregate into an ER. But that would be pretty ugly and inefficient. I'm tempted to make this a utility function (e.g., `downsample_results`) rather than a proper `Filter`, at least for now. Or maybe a standalone class that supports the `fit()` interface but doesn't inherit from `Transformer` (that way it should still work in a `Graph`, even if it doesn't have all of the standard magic that applies to `Stim` inputs).

adelavega:
the latter would be more useful, i think. would allow for graphs, which would be a clean consistent specification of the provenance.

adelavega:
Could this be a built-in function of the `Transformer` or `Extractor` class?

tyarkoni:
Theoretically yes, but I think that would be an inefficient way to do it. You'd need to then specify the target sampling rate to each `Transformer` at initialization, and I don't really like the idea of adding even more logic to the output of the `transform()` call before it gets passed to an `ExtractorResult`. A related possibility would be to do it in the `ExtractorResult` itself, but that suffers from similar issues.

I think having a standalone function or a new `Transformer`-like class specifically for this is the way to go. I don't have a strong preference between the two, but if we went with the former, we could potentially add an optional `sampling_rate` argument to the `Graph`, so that it would implicitly invoke the standalone downsampler after the `merge_results` step.

adelavega:
Okay that works for me. I would just like for the last Extractor element in the history to reflect the extractor itself, versus a downsampling operation (makes it easier to track provenance). 

That approach makes sense to me though.

tyarkoni:
We could still potentially update the `ExtractorResults._history` attribute directly from `ExtractorResults.resample()`, though I have mixed feelings, and some fields (e.g., the name of the transformer) would presumably have to be left blank.

adelavega:
That's fine, I'm actually saying I prefer to see the actual extractor as the last element of the history so this works

tyarkoni:
Oh, sorry, I misread. Then yeah, you're all good!

adelavega:
Closed by #430

PR #430, opened 2020-10-30
Resample ExtractedFeatures

adelavega:
I took a stab at implementing #426 , mostly copying over the logic from pybids.

I decided to combine the upsample and resample logic, although I could see the argument against (although I also find it hard to "detect" the current sampling rate, so it seems easier to simply upsample it). 

Still need to:
 - Do the operation in place (e.g. replace the data in the `ExtractedFeature` object)
 - Consider `object_id` and other cases that could screw this logic up
 - Add to `Graph`

I think otherwise it's fairly close to the logic being implemented by pybids, however, something about the results seem a bit funny to me.

Take this example:
```
from pliers.extractors import RMSExtractor 

ext = RMSExtractor()

res = ext.transform(get_test_data_path() + '/audio/homer.wav')    

In [18]: res.to_df()                                                                                                  
Out[18]: 
    order  duration     onset  object_id       rms
0       0   0.04644  0.000000          0  0.009556
1       1   0.04644  0.046440          0  0.028694
2       2   0.04644  0.092880          0  0.045152
3       3   0.04644  0.139320          0  0.053220
4       4   0.04644  0.185760          0  0.059779
..    ...       ...       ...        ...       ...
69     69   0.04644  3.204354          0  0.020420
70     70   0.04644  3.250794          0  0.016727
71     71   0.04644  3.297234          0  0.007110
72     72   0.04644  3.343673          0  0.004508
73     73   0.04644  3.390113          0  0.004065

In [19]: res.resample(2)                                                                                              
Out[19]: 
   onset  duration       rms
0    0.0       0.5  0.009556
1    0.5       0.5  0.050444
2    1.0       0.5  0.115955
3    1.5       0.5  0.223625
4    2.0       0.5  0.143149
5    2.5       0.5  0.169563
6    3.0       0.5  0.000000
```

It seems that the first sample is in fact only derived from the first value, whereas the last sample is not reflecting all the values ranging from 3-3.5s.

I imagine the most important variable here would be `x_new` which is the output of `np.linspace(0, n - 1, num=num)`

For this example, the value is:

```
ipdb> x_new                                                                                                           
array([   0.        ,  572.66666667, 1145.33333333, 1718.        ,
       2290.66666667, 2863.33333333, 3436.        ])
```

Given that these values reflect the original time in seconds * 1000, I would imagine the first value of `0` is why the first resampled value only reflects the first value. However, given the last value being 3436, I'm surprised the final value is `0` and doesn't reflect the final values from 3-3.5s.

Any ideas? I'll dig in further just taking a break for now, need to understand better now interp1d works.

### DIFF
--- a/pliers/extractors/base.py
+++ b/pliers/extractors/base.py
@@ -2,13 +2,15 @@
 
 from abc import ABCMeta, abstractmethod
 import json
+import math
+from scipy.interpolate import interp1d
 
 import pandas as pd
 import numpy as np
 from pandas.api.types import is_numeric_dtype
 
 from pliers.transformers import Transformer
-from pliers.utils import isgenerator, flatten, listify
+from pliers.utils import isgenerator, flatten
 
 
 class Extractor(Transformer, metaclass=ABCMeta):
@@ -102,9 +104,9 @@ class ExtractorResult:
                 values, the special value 'auto' can be passed, in which case
                 the object_id column will only be inserted if the resulting
                 constant would be non-constant.
-            extractor_params (bool): if True, returns log_attributes of 
+            extractor_params (bool): if True, returns log_attributes of
                 at extraction time, as stored in transformer_params attribute
-                in ExtractorResult.history. These are returned as serialized 
+                in ExtractorResult.history. These are returned as serialized
                 dictionary in extractor_params column.
 
         Returns:
@@ -205,6 +207,59 @@ class ExtractorResult:
     def history(self, history):
         self._history = history
 
+    def resample(self, sampling_rate, in_place=True, kind='linear'):
+        """Resample the ExtractorResult object to the specified sampling rate.
+
+        Parameters
+        ----------
+        sampling_rate (float)
+            Target sampling rate (in Hz).
+        kind : {'linear', 'nearest', 'zero', 'slinear', 'quadratic', 'cubic'}
+            Argument to pass to `scipy.interpolate.interp1d`; indicates
+            the kind of interpolation approach to use. See interp1d docs for
+            valid values. Default is 'linear'.
+        """
+
+        # Cast onsets and durations to milliseconds
+        onsets = np.round(self.onset * 1000).astype(int)
+        durations = np.round(np.array(self.duration) * 1000).astype(int)
+        gcd = np.gcd.reduce(np.r_[onsets, durations])
+        bin_sr = 1000. / gcd
+
+        onsets = np.round(self.onset * int(bin_sr)).astype(int)
+        durations = np.round(np.array(self.duration) * bin_sr).astype(int)
+
+        max_duration = self.onset[-1] + self.duration[-1]
+
+        n = int(math.ceil(bin_sr * max_duration))
+        x = np.arange(n)
+
+        # Calculate final number of samples after re-sampling
+        num = int(np.round((sampling_rate / bin_sr) * n))
+
+        resampled = {}
+        for f_name in self.features:
+            values = self.data[f_name]
+
+            ts = np.zeros(n, dtype=values.dtype)
+            start = 0
+            for i, val in enumerate(values):
+                _onset = int(start + onsets[i])
+                _offset = int(_onset + durations[i])
+                ts[_onset:_offset] = val
+
+            # What does first sample correspond to? Only one
+            f = interp1d(x, ts, kind=kind)
+            x_new = np.linspace(0, n - 1, num=num)
+            resampled[f_name] = f(x_new)
+
+        new_dur = 1 / sampling_rate
+        new_onsets = np.arange(0, math.ceil(max_duration), new_dur)
+        new_onsets = new_onsets[new_onsets < max_duration]
+
+        return pd.DataFrame(
+            {'onset': new_onsets, 'duration': new_dur, **resampled})
+
 
 def merge_results(results, format='wide', timing=True, metadata=True,
                   extractor_names=True, object_id=True, extractor_params=False,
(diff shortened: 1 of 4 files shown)


rbroc:
Hey ale, I have a few remarks, which may explain why the values are off and point to a simpler implementation.

I think the main issue here is that the new onsets in the resampled df are not correct, they are not the time points that resampled values refer to. 

`interp1d` returns the values of the time series at times `x_new`, thus the new onsets should simply be `x_new`.
One more problem here, though, is that the onsets in `x_new` themselves here are not correct. `np.linspace` does not necessarily return the onsets corresponding to the desired sampling rate in Hz - it simply returns `num` equally spaced onsets between `0` and `n-1`. Keeping `num` constant, if you increase `n-1` the interval between values would increase, which is not the desired behavior in our case.

Take your example with `n-1` being `3436`, and `sampling_rate = 2` (Hz). 
Here, `np.linspace` won't return `[0, 500, ..., 3000]` but values at a something like [   0. ,  572.66666667, 1145.33333333, 1718.        , 2290.66666667, 2863.33333333, 3436. ]. 
If `n-1` was 3438 or 3498, the output of `np.linspace` would differ. It'd yield the same number of samples, but different values (with larger spacing).

I think what we're after is actually something different and simpler than the current implementation, i.e., if `r` is the `ExtractorResult`, we'd need to:
- estimate the interval between consecutive onsets as `1/sampling_rate`, which is also going to be the new duration (`new_dur`)
- use `np.arange` to generate the new onsets (`new_onsets`), passing `r.onsets[0]` as start point, `r.onsets[-1]` as end point and `1/sampling_rate` as step.
Then, for each feature in `r`:
- pass `r.onsets` and `r.data[feature]` to `interp1d`.
- apply `interp1d` to `new_onsets` to get the new values.

If this makes sense (but correct me if I'm wrong), the function could be simplified into something like this:

```
def resample(self, sampling_rate, in_place=True, kind='linear'):

        new_dur = 1 / sampling_rate
        x_new = np.arange(self.onset[0], self.onset[-1], new_dur)
        
        resampled = {}
        for f_name in self.features:
            f = interp1d(self.onset, self.data[f_name], kind=kind)
            resampled[f_name] = f(x_new)

        return pd.DataFrame(
            {'onset': x_new, 'duration': new_dur, **resampled})
```

where original and resampled timeseries would look like this.
download


The conversion to milliseconds and upsampling currently done before looping through features and resampling should also be superfluous. 

One more remark on the current implementation: due to the way `new_onsets` are generated, there can be shape mismatches between the number of onsets returned by the function and the number of values (`resampled`), which causes the function to fail (e.g. try using `obama_speech.wav` as stim).

adelavega:
@tyarkoni I dug into this, I think I copied over the pybids logic accurately, and fixed the onsets issue. 

The issue was that without doing filtering of the signal on downsampling (in this instance, a 5th order Butterworth filter), downsampling simply takes the value at the matching timepoint with no interpolation. Interpolation only comes into play if you ask for a value that is not a multiple of the original frequency (and need to interpolate between two neighboring values).

This is true in pybids too, and was true before satra added the filtering logic. 

I re-added the filtering logic, and here's the results from the first 60 seconds of Sherlock, extracting RMS and downsampling to 3hz (from 21.5hz)

image

Black = original at 21.5 hz
Yellow = without filtering
Blue = with butterworth filtering

At @rbroc's suggestion, I then convolved these with an SPM HRF:

image

This would suggest the filtering better represents the original signal after convolution. 

Here's a closer look at the first 5 seconds:

image

The only un-intuitive thing is that the first value (e.g. onset 0, duration 0.3s), doesn't really represent the span of the duration.

Otherwise in terms of pliers, to do:
- Parameterize Butterworth filter?
- Return resampled results as df, or perform in place into `ExtractResult` object.
- Add parameter to `Graph`.

tyarkoni:
Thanks for looking into this; that all sounds good! The convolved, filtered signal looks pretty much dead-on (at least in this case).

adelavega:
Cool. How much parameterize do you think we should allow for the Butterworth filter?

Any comment on returing results as as dataframe or updating the `ExtractorResult` object in place? (The latter seems more "pliers-thonic" probably).

adelavega:
@tyarkoni I was starting to implement in place resampling, and realized that it's impossible to re-create the `data` object for all Extractors, since `to_df` is a one way conversation (can vary across extractors). Thus, I think `resample` should always return a pandas df. 

It might then make more sense to include the option to resample in `to_df`. The downside is that then we have to make `resample` compatible with all variations of `to_df`'s arguments. If we keep it a seperate function we can call `to_df` in the way that makes most sense for that function (e.g. long format, with object_ids etc..).

adelavega:
Also, I don't think I need `resample` to be available in a `Graph`. I'm not using Graphs for extraction in Neuroscout, and I could simply add to Neuroscout to request a resampled_df when I call `to_df` for each `ExtractedResult`.

pliers/extractors/base.py
```diff
@@ -205,6 +207,76 @@ def history(self):
...
+        """Resample the ExtractorResult object to the specified sampling rate
+
+        Parameters
+        ----------
+        sampling_rate (float)
+            Target sampling rate (in Hz).
+        filter_signal: (bool)
+            Apply Butterworth filter to signal prior to resampling
+        filter_N: (int)
+            The other of the Butterworth filter
```
tyarkoni, line 221:
```suggestion
            The order of the Butterworth filter
```

tyarkoni (approved):
LGTM! I only found one clearly very important bug.

adelavega:
Great. I'm going to add tests. I'm going to keep this function as a class method of `Extractor` unless you have a strong objection. If so, could move it to `extractors.misc`, but it seems like a bit of a dark closet to throw it into (might be useful for others).

tyarkoni:
You mean `ExtractorResults`? I don't have strong objections, but it seems a bit odd to have a resampling method implemented on that class return a `DataFrame` rather than another `ExtractorResults` instance (and as you say, there's no good way to do the latter without losing information). So my preference would be to operate strictly on DataFrames (which would also make the function more broadly useful outside the package). If the latter is a pain, then yeah, let's at least move it outside of the class into a separate function, so that it takes an `ExtractorResults` as the first argument.

adelavega:
Okay that sounds good, I think a db util function would be fine. 

Ah, also, I need to think about how to deal with `object_ids`. One option, especially if we're making this a df util function (since there's no guarantee that they deal with all forms of `ExtractorResults`) would be to simply ignore `object_id` for now. Or to `group_by` `object_id`. 

In practice though I'm not sure that any extractors that return `object_ids` make sense to resample (at least for now), and we could cross that bridge when we get to it if there is such a case.

tyarkoni:
Agreed, resampling an extractor with multiple objects would be a bad idea unless you could guarantee that each one is always the same object, in which case we could resample separately. But since there's no such guarantee, and this is something of an edge case, probably best to just issue a warning any time multiple IDs are detected.

adelavega:
okay, done, it is now in `utils.base`. i also made it so it takes the long format df instead, as it's easier to tell which columns are features that way.

once tests pass, i'll merge.

PR #431, opened 2020-11-12
FIX: Fix resample upsampling and add upsampling test case

adelavega:
- Minor bug in #426 such that upsampling failed, due to a rounding error. 
- The requested interpolated values were also subject to a rounding error, so I fixed them to the new onsets * upsampled sr, to ensure they were exactly the same. 
- Fixed and added upsampling test

### DIFF
--- a/pliers/tests/test_utils.py
+++ b/pliers/tests/test_utils.py
@@ -90,10 +90,14 @@ def test_resample():
 
 
     # Test upsample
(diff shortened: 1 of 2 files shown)


adelavega:
@tyarkoni tests should pass soon (unrelated failures). let me know if any objections to merging.

adelavega:
allright, tests passed, merging

PR #432, opened 2020-12-02
MFCC Energy Extractor

adelavega:
WIP. This is a skeleton for Darius to give it a shot if he wants to. Otherwise I'm happy to finish this. @jsmentch


jsmentch:
tagging in @darius522

darius522:
@adelavega Hi Alejandro. Happy to give this a shot. Looks like I don't have the wright permission for that pr yet.

adelavega:
@darius522 I invited you to my fork, hopefully that fixes the issue. Happy to answer any questions you have otherwise. You can also fork my fork, and then do a new PR to here if you want

darius522:
As suggested per the initial extractor skeleton, the implementation extracts the two features by dct'ing the low and high quefrency spectrum. Although MFCCs are commonly used to describe timbre (although a bit less for f0 estimation), I'd like to open the discussion to other alternate approaches, which could potentially improve the accuracy and/or the efficiency of the extractor. 

1. For timbre: [Spectral centroid](https://librosa.org/doc/0.8.0/generated/librosa.feature.spectral_centroid.html), time-domain features such as [zero-crossing rate](http://man.hubwiz.com/docset/LibROSA.docset/Contents/Resources/Documents/generated/librosa.feature.zero_crossing_rate.html) and envelope-derived features could also be considered. Tackling timbral features extraction is generally tricky as it heavily depends on the nature of the signal being analyzed (e.g. there's an [entire library](https://github.com/AudioCommons/timbral_models) dedicated to timbral models that could be helpful to look at moving forward).

2. As for the F0 extraction, algorithms specifically designed to extract the fundamental may potentially return more accurate results (probably to the detriment of efficiency). See [PYin](https://librosa.org/doc/main/generated/librosa.pyin.html). 

Again, putting this out there, mainly as food for thought and potential future improvements. What I like about the current implementation is that it essentially derives two features out of the MFCCs, I'd thus expect the efficiency to be hard to beat.

adelavega:
cc: @PeerHerholz @rbroc

rbroc:
@darius522 thanks for the pointers! I agree with @adelavega - if this is the most reliable pitch extraction methods, as NS extraction takes place offline, I guess efficiency shouldn't be a big concern.

However, [PYin](https://librosa.org/doc/main/generated/librosa.pyin.html) may actually be a cool feature to have in itself, not only wrt to pitch, but also because it seems like we get features coding for voicing/no-voicing probability for free - which could be an especially interesting for speech people. 

Which also reminds me that a while ago I had found some work estimating F1 and F2 formant frequencies (which describe vowel space) from LPC coefficients, which we can now get [via librosa](https://librosa.org/doc/main/generated/librosa.lpc.html). May be worth looking into that soon-ish (adding a reference to this in the gdoc).

One note on the PR: a few months ago we had implemented a [`MetricExtractor`](https://github.com/PsychoinformaticsLab/pliers/blob/master/pliers/extractors/misc.py) with the purpose of "second-order" features - e.g. taking an `ExtractorResult`, converting it to a `SeriesStim` through [`ExtractorResultToSeriesConverter`](https://github.com/PsychoinformaticsLab/pliers/blob/master/pliers/converters/misc.py#L8), and passing it through a custom function (@adelavega, we had used it for BERT surprisal/entropy).

That kind of workflow may actually fit well here.

adelavega:
Hey Darius,
Couple notes:
1) Roberta is going to try using the `MetricExtractor` to get these features, which would allow us to not have to dedicate a new extractor for this. That is, re-use the `MFCC` extractor. It may be too complex for this approach but worth trying
2) I think you make some good points about other potential timbre model, and as far as pliers is concerned that would be useful. However, no need to rush into that, as we need to see if we get any traction at all with even a simple Timbre feature. It's not clear how well we could tease apart the different timbre models in the brain with the current analysis pipeline we have. 

I'll wait for roberta to try this out w/ `MetricExtractor` and if its not possible with that approach I'll circle back to this and give it a review.

darius522:
@adelavega @rbroc thanks for providing additional context on the issue. Based on Roberta's comment on the MetricExtractor, it sounds like the class is perfectly suited to this scenario, as the timbre and pitch are currently being derived directly from a "first-order" feature (MFCCs). Additionally, if we're also thinking about implementing a more robust extractor specifically for the fundamental frequency but also for timbre, it may be worth thinking about either renaming these MFCC-based features (maybe "low energy" for pitch and "high energy" for timbre ?) or simply considering not including them. But again, I may not have the proper background to judge which approach may be the most beneficial from a neuroscience perspective.

Would be more than happy to look into the formant feature extractor later down the road, if needed!

rbroc:
alright, looked into it, and seems like `ExtractorResultToSeriesConverter` + `MetricExtractor` is **not** a viable option for this scenario. The converter groups values from an `ExtractorResult` by `onset`, not by feature - i.e. it returns one `SeriesStim` per onset. The `MetricExtractor` thus applies a function to all values for a given onset across features - rather than to the series of all values for a certain feature across onsets.

Unless we want to go down the road of implementing Extractors/Converters supporting operations over time series (which could be cool by the way, not only for this case, but also for other stuffy like autoregressive features), then we actually need a specific extractors for pitch/timbre. 

Will look into whether LPC-based formant features is actually a viable options - right now I can only find Matlab code for it https://se.mathworks.com/help/signal/ug/formant-estimation-with-lpc-coefficients.html 

Thanks for this, @darius522 !

pliers/extractors/audio.py
```diff
@@ -618,3 +618,54 @@ def _extract(self, stim):
         return ExtractorResult(preds, stim, self, features=labels,
                                onsets=onsets, durations=[dur]*len(onsets),
                                orders=list(range(len(onsets))))
+
+
+class TimbrePitchExtractor(AudioExtractor):
```
adelavega, line 623: Let's go ahead and rename this extractor something like `MFCCEnergyExtractor`

Darius, suggestions are welcome

pliers/extractors/audio.py
```diff
@@ -618,3 +618,54 @@ def _extract(self, stim):
...
+        self.n_mfcc = n_mfcc
+        self.n_coefs = n_coefs
+        self.hop_length = hop_length
+        self.librosa_kwargs = librosa_kwargs
+        super().__init__()
+
+    def _extract(self, stim):
+
+        #https://librosa.org/doc/main/generated/librosa.feature.mfcc.html
+        mfccs = librosa.feature.mfcc(y=stim.data,
```
adelavega, line 645: I think our only concern is that we already have an MFCC extractor and it "feels bad" to not re-use it properly.

@tyarkoni does it bother you to "re-use" MFCC here. We could also make this extractor more general way to extract groups of coefficients not just "high and low". I also don't want to over-think this though and rather just merge and iterate later if needed. 

FWIW, I think the approach of taking the top 13 components to extract high and low energy is a pretty common thing to do in audio feature extraction.

e.g.:

"The components of MFCCs are the first few DCT coefficients that describe the coarse spectral shape. The first DCT coefficient represents the average power in the spectrum. The second coefficient approximates the broad shape of the spectrum and is related to the spectral centroid. The higher-order coefficients represent finer spectral details (e.g., pitch). In practice, the first 8–13 MFCC coefficients are used to represent the shape of the spectrum. "
https://www.sciencedirect.com/topics/computer-science/cepstral-coefficient

pliers/extractors/audio.py
```diff
@@ -618,3 +618,54 @@ def _extract(self, stim):
...
+            norm='ortho', 
+            overwrite_x=False)
+
+        lq_mfs = 10 ** (lq_mfs / 20.)
+        hq_mfs = 10 ** (hq_mfs / 20.)
+
+        out_val = {'pitch':lq_mfs,'timbre':hq_mfs}
+
+        return ExtractorResult(out_val, stim, self,
+                               features=['pitch','timbre'])
```
adelavega, line 671: Makes sense to me to rename the feature "high_mfs" and "low_mfs" or something like that.

adelavega (commented):
@darius522 can you also add tests? see similar audio extractors for tests, and let me know if you need help.

pliers/extractors/audio.py
tyarkoni: Fine with me to call it internally instead of invoking another extractor, seeing as it's just one call. I think it might even be preferable this way—we don't need all of the standard logging machinery to kick in.

pliers/extractors/audio.py
tyarkoni: At some point we might need/want to move all of the librosa-based classes into their own module and/or rename them all to make the extractor source explicit. But we can hold off until it becomes an issue (e.g., if someone wants to use a different tool to extract a similar feature).

tyarkoni (approved):

pliers/extractors/audio.py
```diff
@@ -618,3 +618,55 @@ def _extract(self, stim):
...
+            n=self.n_mfcc, 
+            axis=- 1, 
+            norm='ortho', 
+            overwrite_x=False)
+
+        low_mfs = 10 ** (low_mfs / 20.)
+        high_mfs = 10 ** (high_mfs / 20.)
+
+        out_val = {'low':low_mfs,'high':high_mfs}
+
```
adelavega, line 670: To allow the pliers `to_df` method to behave appropriately we need to define a `get_feature_names` function. 

This is an example (I have not verified this code);

```suggestion
def get_feature_names():
    return [f"{name}_{i}" for name in ['low', 'high'] for i in range(0, self.n_n_mfcc)]   
```
if `n_mfcc` it should return the following:
```
['low_0',
 'low_1',
 'low_2',
 'low_3',
 'low_4',
 'low_5',
 'low_6',
 'low_7',
 'low_8',
 'low_9',
 'high_0',
 'high_1',
 'high_2',
 'high_3',
 'high_4',
 'high_5',
 'high_6',
 'high_7',
 'high_8',
 'high_9']
```
Thus, the 2-D matrix that is return to `ExtractorResult` should be (number of frames) x (n_mfcc * 2) (for each low and high).

adelavega:
@darius522 if you have any updates, let me know I might have a little bit of time to give feedback, if not not worries and have a nice winter break!

darius522:
@adelavega just pushed my proposed changes to your fork. Taking into consideration the discussion above and after giving this a bit more thinking, I've decided to approach the matter a bit differently. From my perspective it was a bit incoherent to have some extractors extracting "first-order" features while some others extracting the same "first-order" but processed (sounded a bit redundant as well).

I've added a class method to the `LibrosaFeatureExtractor `class called `_proc_func` which aims to be overridden by a subclass in case the subclassed extractor aims at processing its first-order features further. In our case, the `MFCCExtractor` does NOT override this method, while the MFCCEnergyExtractor class, which inherits from `MFCCExtractor`, does. It's a bit more reusable, clear, and flexible this way, in my view.

To overcome the recent issues related to output shape, instead of having one single extractor outputting "high_mfs" and "low_mfs", I've split them into two extractors (`MFCCLowEnergyExtractor` and `MFCCHighEnergyExtractor`). Beside resulting in an output shape consistant with the other extractors', this also gives the freedom to users to extract the low and high energy with arbitrary bounds, independently from each other.

This is still WIP but If this suggestion is acceptable to you guys, I can finalize everything next (including the test cases).

pliers/extractors/audio.py
```diff
@@ -618,3 +634,67 @@ def _extract(self, stim):
...
+    def __init__(self, n_mfcc=48, n_coefs=35, hop_length=1024, **librosa_kwargs):
+        self.n_mfcc = n_mfcc
+        self.n_coefs = n_coefs
+        self.hop_length = hop_length
+        self.librosa_kwargs = librosa_kwargs
+        super().__init__(n_mfcc=n_mfcc)
+
+    @override
+    def _proc_func(self,stim):
```
adelavega, line 691: I see where your going with this, but I'd like to avoid using "custom" ways of over-riding functionality, especially the `override` decorator (since this is not "standard" python). The main reason I say this is because this will reduce maintenance burdens in the future since it will be more obvious what this is doing, if it is done in a more "standard" way.

I think you should be able to accomplish this same behavior using `super()`. It seems that  you should be able to override the `_get_values` function with the contact of this `_proc_func`. Instead of calling `self._get_values` you would call `super()._get_values` to get the MFCCs, then post-processes them further and return the energy extracted from the MFCCs.

I haven't thought it through completely but I believe this should let you do what you're doing here, without manually keeping track of whether a function as been over-ridden or not. `_get_values` should simply return the values of whatever the current extractor is.

adelavega, line 691:
```suggestion
    def _get_values(self,stim):
```

pliers/extractors/audio.py
```diff
@@ -618,3 +634,67 @@ def _extract(self, stim):
...
+        vals = self._get_values(stim)
```
adelavega, line 692:
```suggestion
        vals = super()._get_values(stim)
```

pliers/extractors/audio.py
```diff
@@ -618,3 +634,67 @@ def _extract(self, stim):
...
+        low_mfs = fft.dct(np.transpose(vals[:self.n_coefs]),
```
adelavega, line 661: I think I prefer only having a single "EnergyExtractor". It seems too redundant to me to duplicate the logic of the low and high extractors. If we ever had to make any changes to this logic, we'd have to do it twice.

The only difference between the two extractors is this single line. 

How about instead, we add a parameter (`direction`?) that dedicates if we are taking `n_coefs` from the `high` or `low` end of the spectrum. If the user wants both, they simply run the same extractor twice with different parameters.

adelavega (commented):
Darius, I like where your going with this PR but I added some feedback to make this approach more Pythonic. Otherwise, I think the approach of trying to reduce "re-doing" the MFCC call is a good one. I just don't want to overly complicate the generic logic of the `AudioExtractor` just for that.

I think what I've proposed should allow you to accomplish what you're doing here in a more "normal" "Pythonic" way that will be more easy to debug in the future.

adelavega:
@PeerHerholz can you give this a try? I can't remember exactly what the output we should expect looks like, (its n_mfcc dimensional right?) but I think this is pretty much where it needs to be.

PeerHerholz:
Check @adelavega, I'm on it!

PeerHerholz:
KK, I installed `pliers` from this commit within a new `conda env` and I tested the `class` on an example and compared it to the functions I have used so far. Based on that I think we would need to check a few things again. The respective gist is [here](https://gist.github.com/PeerHerholz/662ba08c15e7c4a6dfb3eb8eff431cfe). Could y'all (@adelavega, @jsmentch  and @rbroc) have a look?

adelavega:
I just looked at it quickly, thanks. So do you know what the problem is?

Darius implemented the DCT logic and wrote some tests for it (fixing pliers tests in #457).

adelavega:
This is pretty much the entirety of the logic: 

```
    def _get_values(self,stim):
        # This super function calls MFCCExtractor with the arguments provided to this class
        vals = super()._get_values(stim)

       # From here forward we "post process" the MFCCs to look at the low/high
        subset = vals[:self.n_coefs] if self.direction == 'low' \
            else vals[(self.n_mfcc-self.n_coefs):]

        subset_mfs = fft.dct(np.transpose(subset), 
            type=2, 
            n=self.n_mfcc, 
            axis=- 1, 
            norm='ortho', 
            overwrite_x=False)
        subset_mfs = 10 ** (subset_mfs / 20.)

        return subset_mfs.T
```

The shape of the results at least looked correct right?

adelavega:
@PeerHerholz I noticed you are using `mfs = feature.melspectrogram` not `mfs = feature.mfcc`

jsmentch:
So it looks like direction='low' and 'high' both return the low quefrencies but those low coefs look good? The logic looks correct to me so far though

adelavega:
Allright, I think I fixed the pliers issues (caching was messing up the direction argument).

Now the issue is the results from this are similar, but not identical to that from @mih 's studyforrest analysis.

Using an example audio file, here is the vizualization of pitch using Michael's code:

image

If we take Michael's code and use librosas's mfcc (instead of using custom DCT_II to compute MFCCs from librosas' melspectrogram), this is what we get: 

image

Finally, if we also use scipy's DCT instead of Michaels to compute pitch from the MFCCs (as implemented here), this is what we get:

image

My guess is there's two things going on:

1) Michael add some normalization logic I don't fully understand in the DCT step. This is probably different than what scipy does,
    I'm also not fully confident in how we're calling DCT. @darius522 wrote that code maybe he can explain

2) librosa does some additional preprocessing when computing MFCCs as seen here:

```
    if S is None:
        S = power_to_db(melspectrogram(y=y, sr=sr, **kwargs))

    M = scipy.fftpack.dct(S, axis=0, type=dct_type, norm=norm)[:n_mfcc]
```

Anyone with more knowledge about this want to chime in? @satra @mih? Thank  you!

satra:
@adelavega - can we not call this pitch ? :) are you truly trying to extract pitch? i don't know of any general purpose pitch extractors. 

but in general it looks like you are simply trying to get at a mel spectrum of a subband of mfccs.  in librosa you should be able to do: mfcc with either concurrent liftering (emphasize certain coefficients) or subsequent band removal (set the coefficients to zero) followed by mfcc to mel transform (it has a function for it.

let me know what you are trying to get at, since it's not really pitch :)

adelavega:
I'm only calling in pitch informally-- in the extractor its just "High" or "Low" energy

I'm trying to replicate what Michael did in this paper:
https://f1000research.com/articles/4-174/v1
"Low-Quefrency and High-Quefrency Mel-Frequency Spectrum (lq_mfs, hq_mfs)"

Also, tell that to @mih ;):
"These two sets of features represent broad-spectrum information (timbre) and fine-scale spectral structure (pitch) respectively."

but maybe we should talk and see if there are other more "interesting" features to prioritize.

satra:
try the mfcc -> zero out the relevant mfcc coefficients based on high/low -> mfcc_to_mel (make sure the dct_type, norm, and number of mel filter banks all match)

> "These two sets of features represent broad-spectrum information (timbre) and fine-scale spectral structure (pitch) respectively."

those are very qualitative associations :) 

and yes we can talk about what features to consider in the general space of audio. for example, you can also run a hilbert transform decomposition of the signal which will break out envelope and fine structure.

PeerHerholz:
thx for the additional input @satra, we should definitely note that these are qualitative associations.

> try the mfcc -> zero out the relevant mfcc coefficients based on high/low -> mfcc_to_mel (make sure the dct_type, norm, and number of mel filter banks all match)

I think that should happen within the code adapted from @mih as outlined within this [gist](https://gist.github.com/PeerHerholz/662ba08c15e7c4a6dfb3eb8eff431cfe).

satra:
@PeerHerholz - have you tried just using the librosa functions? 

https://librosa.org/doc/0.7.1/generated/librosa.feature.mfcc.html
in between just zero out the relevant coefficients.
https://librosa.org/doc/0.7.1/generated/librosa.feature.inverse.mfcc_to_mel.html

PeerHerholz:
IIRC no. So far, I have always used the code from the `studyforrest` gang. @adelavega tried the `mfcc` but not the `mffc_to_mel`.  Should we have a look at that @adelavega? Happy to run that tomorrow.

adelavega:
I think that's conceptually what I'm doing, but instead of zero-ing out, I used a subset of MFCCs, and used `fft.dct` instead of `mfcc_to_mel`. Should be a straight forward change I think.

adelavega:
Allright I've implemented it and the same audio file this is what I get.

For "low":
image


and for "high":
image


Looks reasonable to me, I suppose? 

For fMRI one question is how many mels is reasonable? What's the least we can go with (easier models to run) that is still useful?

satra:
@adelavega - thanks. for the high quefrencies the image somewhat makes sense, for the low quefrencies it looks way too low frequency. possible. 

- is the signal sampled at 22KHz?
- are the spectrograms showing values in dB

regarding mel bins 128 (which i think is the default) should be fine.

pliers/tests/extractors/test_audio_extractors.py
```diff
@@ -29,8 +29,13 @@
                                BeatTrackExtractor,
                                HarmonicExtractor,
                                PercussiveExtractor,
+<<<<<<< HEAD
+                               AudiosetLabelExtractor,
+                               MFCCEnergyExtractor)
+=======
```
satra, line 35:
```suggestion
                               MFCCEnergyExtractor,
```

pliers/tests/extractors/test_audio_extractors.py
```diff
@@ -29,8 +29,13 @@
...
                                AudiosetLabelExtractor
                                )
+>>>>>>> origin/master
```
satra, line 38:
```suggestion
```

pliers/tests/extractors/test_audio_extractors.py
adelavega: oh, weird

adelavega:
> * is the signal sampled at 22KHz?

44100 Hz it seems

> * are the spectrograms showing values in dB
No, it's in power. I think the `mfcc_to_mel` function applies a dB to power transformation

adelavega:
@satra could it be that the values are normalized?

Again, this is the output from @mih's code: 

Low: 
image
 and

High:
image

Looks plausible to me, but maybe we should try another audio file as well.

adelavega:
@satra and I looked at this closely and conceptually it looks good.

Looks like most likely the inverse DCT in librosa is different from Michaels. For type II DCT you need to do a type III inverse DCT. 
There may some normalization differences as well. 

I'm going to do another check vs's Michael's code to ensure this is where it diffiers, but the librosa implementation should be truthworthy

Other notes:

- [ ] Capture all arguments needed for `mfcc_to_mel` in the extractor
- [ ] All others pass to MFCC as kwargs

adelavega:
@satra I did a close comparison, taking the log in this Extract but straight in a notebook to compare side by side: https://github.com/adelavega/mfcc-test/blob/master/timbre_pitch_test.ipynb

The implementation here is results in pretty similar results to Michael's but not identical.
I think it comes down to the fact that if you look at the output of librosas's MFCC, it is not equivalent to the MFCC's that M's code produces. In principle they are similar, but whatever difference in normalization or DCT is occuring, means they are not exactly the same.

Regardless, I think we agreed last time the results are close enough, and in principle correct, so we should trust librosa's implemenation, so I'll charge forward with this and add some tests.

satra:
@PeerHerholz, @mih  - do you have a pointer to the original implementation? a transform to mfcc requires a DCT and the conversion from mfcc to mel requires an inverse dct. if the first one is a standard dct type-II (with ortho), it's inverse is equivalent to a dct type-III (with ortho), but these are two different functions. for more details see the descriptions for type II and type III dct here (https://docs.scipy.org/doc/scipy/reference/generated/scipy.fftpack.dct.html)

@adelavega - i agree that we can stick with the librosa implementaion. the code LGTM.

mih:
Hey, the feature extraction was contributed to the project. The original code is here: https://datapub.fz-juelich.de/studyforrest/forrest_gump/stimulus/task002/make_audiofeatures.py

PR #433, opened 2020-12-07
add tensorflow hub extractor

### LINKED ISSUE #428: support TF Hub models
We could consider adding model extractors that pull DL models from Tensorflow Hub (https://www.tensorflow.org/hub). Right now, through keras applications we only have support for object recognition models, but there's interesting stuff coming out in the audio domain (https://tfhub.dev/google/spice/2).

It may be possible to have one extractor for each task (e.g. one class for all audio embedding models), need to look into whether input and output format allow it. 

As an aside, YAMNET is now on TfHub, so we may consider updating our AudioSet extractor so to pull the model from there: https://tfhub.dev/google/yamnet/1

rbroc:
First draft of a (hierarchy of?) Tensorflow Hub extractor(s).
In current implementation, a generic TFHub extractor is implemented which is agnostic to stimulus type + two subclasses for _embedding_ models and _classification_ models - at the moment, these are pretty much the only two task types for which input and output seem fairly standardized (see https://www.tensorflow.org/hub/common_saved_model_apis/images and https://www.tensorflow.org/hub/common_saved_model_apis/text).

Closes #428.


rbroc:
@adelavega @tyarkoni the current implementation could be one way to go about this. 

There's a generic extractor (TFHubExtractor) where you can in principle pass whatever TFHub model, it will pack the output into a number of feature columns with generic names or into a columns with custom names if you pass them (if only one name is passed, it packs everything into a column). It's a class that allows you to use any model, but at your own risk as it remains rather abstract in terms of input type, input preprocessing and output postprocessing.

Then, there's are two modality-specific extractors, one for `ImageStim` and one for `TextStim`, for which input format in TFHub is more standardized (see https://www.tensorflow.org/hub/common_saved_model_apis and following sections). The modality-specific extractors do the type of preprocessing needed to pass the data from each `Stim` type to the models, and the postprocessing needed to pass the output to the `ExtractorResult` object.

This should work for most image classification and embedding models and text embedding models. In terms of output, these extractors are also pretty flexible. If you do not pass any labels/feature names, the extractor will just split the output in a number of features/columns with generic names (`'feature_0', 'feature_1', ..., 'feature_n'`) according to the dimensionality of the model output. If you pass feature names, it will use those. If you pass a single feature name (e.g. `'embedding'`), it will pack everything into one feature/column with that name.

We could also decide to be more specific and *always* constrain the dimensionality of the input to `n_classes` for classification, and to one feature (an embedding vector) for embedding models. We could do so by creating subclasses of `TFHubImageExtractor` for classification and embedding tasks - where all we do is forcing the feature name attribute to be an `n_classes`-long list (of either generic labels or user-specified labels) in the classification case, and one-dimensional feature list (e.g. `['embedding']`) in the embedding case. I think all methods could be inherited as-is from the parent classes.

Let me know what you think - I'll go ahead and write tests if you like the current approach.

rbroc:
one more note: right now the generic extractor crashes (`TypeError: Can't instantiate abstract class TFHubExtractor with abstract methods _input_type`) because `_input_type` is not specified, even when one specifies `_optional_input_type` (which in my understanding would be the most suitable thing here, as it'd make it possible to accept a any `Stim` depending on which model you pass).

Is this expected behavior? In my understanding, specifying `_optional_input_type` should override the need to specify `_input_type`.

If, though, this *is* expected behavior and it's me misunderstanding something - shall I create modality-specific classes for audio and video where `_input_type` is set? Or are there other ways to allow the generic extractor to accept any input, if we want to go for this option?

pliers/extractors/models.py
```diff
@@ -1,14 +1,193 @@
...
+        task (str): model task/domain identifier
+        features (optional): list of labels (for classification) 
+            or other feature names. If a one-element list or 
+            a string is passed, packs output vectors into one column.
+            If not specified, returns numbered features 
+            (feature_0, feature_1, ... ,feature_n).
+        kwargs (dict): arguments to hub.KerasLayer call
+    '''
+
+    _log_attributes = ('url_or_path', 'features', 'transform_out')
```
tyarkoni, line 36: This is one of the problems with using generic classes: we lose the ability to easily log potentially important initialization parameters. I don't know if there's an easy solution to this... if `kwargs` can only contain fairly simple, hashable types, we could just add that to `_log_attributes`. But I'm guessing it's not that predictable, in which case I guess we just have to live with it.

pliers/extractors/models.py
```diff
@@ -1,14 +1,193 @@
...
+
+    def __init__(self, url_or_path, features=None, task=None, 
+                 transform_out=None, **kwargs):
+        verify_dependencies(['tensorflow', 'tensorflow_hub',
```
tyarkoni, line 40: Are all of these dependencies needed here? At least from the code, it looks like it might only be `tensorflow_hub` (for `KerasLayer`).

tyarkoni (approved):
Very minor comments I leave to your discretion, otherwise, looks great!

pliers/extractors/models.py
```diff
@@ -1,14 +1,193 @@
...
+
+
+class TFHubExtractor(Extractor):
+
+    ''' A generic class for Tensorflow Hub extractors 
+    Args:
+        url_or_path (str): url or path to TFHub model. You can
+            browse models at https://tfhub.dev/.
+        task (str): model task/domain identifier
+        features (optional): list of labels (for classification)
```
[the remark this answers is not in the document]
rbroc: added more explanation in new commits

pliers/extractors/models.py
rbroc: only hub is required, indeed.
Removed tensorflow dependency, and moved tensorflow_text to the text extractor.
It is only needed for some models and there's no way to know if it is needed until the model is called, so I've added a warning at initialization when import fails.

pliers/extractors/models.py
rbroc: indeed, arguments to `KerasLayer ` can be pretty much anything depending on the model, and some of it is non-hashable (e.g. the "arguments" dictionary: https://www.tensorflow.org/hub/api_docs/python/hub/KerasLayer). Would lean towards leaving it as it is, but I can add a warning if kwargs are passed if needed.

rbroc:
@tyarkoni @adelavega I've added test and implemented Tal's suggestions so is ready for new review. Tests pass locally but they are not triggered automatically on Travis, and I'm not sure if I can start them manually and how. As for the BERT extractor, I expect these tests may cause some memory issues on Travis (each extractor is tested on more than one model) - but let's see what happens.

By the way, there is probably also some updating to do to the BERT extractors, I can take care of that in a separate PR.

pliers/extractors/models.py
```diff
@@ -1,14 +1,242 @@
...
+                 **kwargs):
+        verify_dependencies(['tensorflow_hub'])
+        self.model = hub.KerasLayer(url_or_path, **kwargs)
+        self.url_or_path = url_or_path
+        self.features = listify(features)
+        self.transform_out = transform_out
+        self.transform_inp = transform_inp
+        super().__init__()
+
+    def get_feature_names(self, out):
```
tyarkoni, line 74: This is currently publicly exposed, so we might want to move the check for `self.features` inside here (instead of doing it in `_extract`) and use that if available. Otherwise a user might naively call `get_feature_names` expecting to get the stored feature names, and instead they'll get the naive enumeration of `feature_*`.

Alternatively, if it's not meant to be public, maybe rename to `_get_feature_names`.

tyarkoni (commented):
Looks good, just one minor suggestion.

pliers/extractors/models.py
rbroc: thanks for the feedback and suggestion, Tal! Moved check for `self.features` inside `get_feature_names`.

rbroc:
tests are failing - probably an issue with pandas version (which we are pinning to `~=0.25` 😬), looking into that asap

rbroc:
@tyarkoni @adelavega I've merged the pandas PR here and tests initially failed with (I think) the same out of memory issues which we've previously encountered for BERT extractors. As a reminder, what happens is not that the whole thing crashes *while* running TFHub extractor tests, but that memory consumption accumulates over tests up to the point where - if there are many heavy model - the error occurs (e.g., in this case, tests don't crash while testing TFHub extractors, but afterwards, when we test BERT extractors).

As a workaround, I moved the test for one of the heavier TFHub models (Electra) in a separate function marked as high-memory and skipped on CI. This seems to be enough to avoid the OOM error and make tests run smoothly. The downside of this is that we are not testing any of the TFHub text models that operate in two steps (downloading tokenizer and tokenizing the input first, then passing the tokenized input to the model of interest). I guess we could in principle live with that (as we do for some of the BERT extractors) - provided that we test locally every now and then.

Alternatively, I don't know the details of how cron works in this context, but we could maybe set up jobs running only high-memory tests, potentially even running less frequently than the rest of the test suite.

In any case, if you approve of the current solution, this may be ready to merge after having merged the pandas fix.

tyarkoni:
Hmm... this could be an architectural problem with our tests. One possibility is that we have module- or package-level pytest fixtures we're binding large objects to, and those get carried through. If so, the solution would be to explicitly delete objects we don't need, or to break up the fixtures. At minimum you can try explicitly calling `del` on extractors after we're done with them in the test, in the event that there are some references to them that are hanging around for unclear reasons.

rbroc:
tried to call `del` but to no effect, sadly. 
Happy to explore other options, if you have any suggestions?

tyarkoni:
Some other thoughts:
* Verify that transformer caching is turned off. We currently do that in `conftest.py`, but it's worth making sure that that setting is sticking in the test modules when pliers is re-imported. If it isn't, that could be the problem.
* It's possible deleting the transformer instances isn't destroying the (memory-intensive) TF models if there's some other reference somewhere. You could try explicit `del ext.model` calls (where `ext` is the extractor). I assume TF is running on a CPU in the tests and isn't automatically getting executed on a GPU; if the latter, you might have to explicitly free up GPU memory in some other way.
* Try moving expensive tests into their own modules, in the event that the issue is coming from the TF layer but isn't due to the model objects sitting around.

rbroc:
awesome, thanks tal! Will try that out tomorrow! 🤞

adelavega:
@rbroc I think when running using forked mode, the pytest print out becomes very ugly and it looks like it hangs, but it doesn't.
Happened to me as well.

It looks like there is one failure in 3.7 and 3.8 which is:
worker 'gw1' crashed while running 'pliers/tests/extractors/test_model_extractors.py::test_tfhub_text'
It's possible that single extractor is using way too much RAM and crashing regardless.

Some suggestions:
- remove `-n auto` from the pytest call. That might make the printout nicer and less likely two memory intensive things run at once
- Mark `test_tfhub_text` to skip unless a variable such as `RUN_HIGH_MEM` = True. Then you could set that variable to True after the rest of the tests run and run in its own line. If that still doesn't work then it may have to be skipped alltogether on CI.

adelavega:
Looks like `-n auto` works when the forked and non forked tests are called separately. This takes the total execution time from ~28-21 minutes, with the first tests completing in only about 12-14 mins (Python 3.6 takes way less for some odd reason). 

@rbroc it's a bit out of the context of this PR so perhaps could merge this and deal with this later, but only thing left is to change the `MetricExtractor` test to use a more basic extractor.

rbroc:
thank you, @adelavega! 🙏 I've opened a separate issue for `MetricExtractor` and will look into it asap. 
Tests are passing, so I assume this one is ready to merge?

adelavega:
Sounds good, let's merge!

PR #436, opened 2021-01-07
Convert CI to Github Actions

adelavega:
Travis CI is not cool anymore, so moving to Github Actions.
Also: Merged @rbroc's PR that fixes HuggingFace's BERT extractors, fixed AudioSet model download, and various minor test regressions.
Oh, also I'm switching to codecov because Coveralls GH action doesn't play nice w/ pytest apparently.


adelavega:
Looks like coverage didnt work, so need to fix that first

adelavega:
Codecov is working!

tyarkoni (approved):
Looks good, thanks! Only suggestion is to maybe add Python 3.8 and 3.9 to the matrix. (We could probably also explicitly update minimum Python version to 3.6 given that we no longer test on 3.6 and 3.5 has now reached its end of life).

tyarkoni:
One minor issue is that if we re-enable the paid tests, they'll all fail because none of the environment variables are set on GitHub. I'll need to look into this at some point and add all the respective secrets, but I imagine it's pretty straightforward.

adelavega:
Yep, should be easy to add those under the repo Settings -> Secrets. 

I'll change the matrix to 3.6-3.9 and merge once passing.

PR #439, opened 2021-01-14
Fix pytest memory leak by running forked

adelavega:
I looked into the memory leak issue and no matter how many things I tried to delete from memory, there was always significant number of leakage from deep learning models, especially. I tried deleting objects, and modules from `sys.modules` but the problem persisted. 

The majority of leakage was occurring in about 8-10 tests that leaked between 200-1200MB each.

Turns out it's quite difficult to clean up objects in memory that are nested within libraries, without accidentally unloading things from pliers or pytest.

Instead, I found a library called `pytest-forked` which allows you to run each test in its own forked process. I tagged all the high memory leak tests as to be forked. Therefore, all imports are done independently and in an isolated fashion. Combined with running in parallel across threads, the tests are finishing in about 8-12 mins, including all tests we previously skipped due to "high memory usage".

### DIFF
--- a/.github/workflows/python-package.yml
+++ b/.github/workflows/python-package.yml
@@ -43,5 +43,5 @@ jobs:
         python -m pliers.support.setup_yamnet
     - name: Test with pytest
       run: |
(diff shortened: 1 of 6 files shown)


tyarkoni (approved):
Oh, sweet, thanks! LGTM.

PR #452, opened 2021-03-20
[CI] Build docker image with github actions.

jdkent:
using github actions to build/distribute docker image

`secrets.CR_PAT` needs to be created by someone with admin privileges.
It's a personal access token, [with package permissions](https://docs.github.com/en/packages/learn-github-packages/about-github-packages#authenticating-to-github-packages), I can do it (if you think I _should_ have admin access, hopefully to just this repository, but maybe need someone that is admin to the organization).


jdkent:
finally got it passing, for some reason I cannot run the `pliers/tests/extractors/test_model_extractors.py` tests, (it times out), but should be ready for review otherwise @adelavega

adelavega:
Ah, that's a common issue with the timeout. This is why in the past I had to split the tests into multiple lines. It probably runs out of RAM.

adelavega:
Nice, looks good. Can you update the README or docs to reference how to use the Docker image?

jdkent:
@adelavega there is a urllib error for the python tests that I'm not sure about, do you have any thought?

jdkent:
@adelavega, I think this is finally ready, there is a soft limit of 14GB of storage on the runner, so I had to delete some very large unused files on the virtual machine so I could build the very large pliers image.

jdkent:
@adelavega, does this pull request look okay? once merged, it should fix the broken tests in #454

adelavega (approved):
Looks excellent. Let's merge!

PR #454, opened 2021-04-06
[DOC] Add video

jdkent:
Adds @adelavega's excellent video explaining pliers at the **TOP**, since people like watching things more than reading, and link corresponding tutorial/course in readme.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -4,6 +4,12 @@
 |PyPI version fury.io| |pytest| |Coverage Status|
 |DOI:10.1145/3097983.3098075|
 
+
+.. raw:: html
+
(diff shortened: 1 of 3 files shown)


adelavega:
As flattered as I am James, maybe the link to the tutorial is sufficient? Only because this is domain specific to fmri/neuroimaging and in theory pliers is a domain general tool.

jdkent:
I think it's still good to have the video directly linked (I think it's a good sign that shows project vitality, investment, and people sometimes prefer to watch something versus read), but I conceded and put the video in Documentation with a disclaimer that pliers is a general-purpose tool.

If the video becomes too out of date we can remove it (or replace it?)

How does that sound?

jdkent:
for whatever reason, the doc build now requires pytest, I added pytest to the doc requirements and refactored setup.py so that the requirements in setup.py will be the same as the requirements in the files.

ready for review @adelavega

adelavega (approved):
LGTM

PR #455, opened 2021-04-14
Change excel engine

### LINKED ISSUE #449: Opening excel files in DictionaryExtractor not compatible with newer pandas version
```
/usr/local/lib/python3.6/site-packages/pandas/io/excel.py in __init__(self, io, **kwds)
    367
    368         if engine is not None and engine != 'xlrd':
--> 369             raise ValueError("Unknown engine: {engine}".format(engine=engine))
    370
    371         # If io is a url, want to keep the data as bytes so can't pass

ValueError: Unknown engine: openpyxl.
```

Looks like we need to change the engine to `xlrd` to fix this issue

adelavega:
Closes #449

### DIFF
--- a/optional-dependencies.txt
+++ b/optional-dependencies.txt
@@ -9,7 +9,6 @@ librosa>=0.6.3
 numba<=0.48
 matplotlib
 opencv-python
-openpyxl
 pathos
 pygraphviz
 pysrt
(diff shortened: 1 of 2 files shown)


adelavega:
Actually, xlrd does not support xlsx files, and doing an upgrade of pandas and libraries made it so that xlsx worked again...

Worked with pandas == `1.1.5`

PR #459, opened 2021-09-09
Parameterize dtype for TFHubExtractor, add ImageResizingFilter + ImageRescalingFilter

anibalsolon:
Using the TFHub module, I've found some issues in the preprocessing part of the images, before injecting them in the model. This PR is a suggestion on how to fix the issues.

One issue is that models require the images to be in the [0, 255] range, using dtype e.g. tf.uint8. My proposal is to let the dtype to be parametrizable, allowing the user to not rescale & keep as uint8.

The other issue is that the specialization of the `_preprocess` function in the `TFHubImageExtractor` class supersede the `transform_inp` custom function. It is an interesting functionality, as it allows to work with the data already in tensors.

Please let me know your thoughts about it!

_____

Edit by @adelavega:

 - `TFHubExtractor`  lets dtype be parameterize
 - To enable users to rescale or reshape inputs, `ImageResizingFilter` and `ImageRescalingFilter` were implemented
 - Exception handling in case model instantiation fails to hint to users to use the appropriate filter.


adelavega:
Cool, I'll try to take a look this but if I don't feel free to bug me

pliers/extractors/models.py
```diff
@@ -154,10 +156,21 @@ def _preprocess(self, stim):
             x = resizer.transform(stim).data
         else:
             x = stim.data
+
         if self.rescale_rgb:
+            if self.input_dtype.dtype.is_integer:
+                logging.warning('Rescaling the input from [0, 255] to [0, 1] and settings '
```
adelavega, line 162:
```suggestion
                logging.warning('Rescaling the input from [0, 255] to [0, 1] and setting '
```

pliers/extractors/models.py
```diff
@@ -154,10 +156,21 @@ def _preprocess(self, stim):
...
+                                'input_dtype to an integer type may result in loss of precision.'
+                                'Please, set rescale_rgb to False.')
+
             x = x / 255.0
-        x = tf.convert_to_tensor(x, dtype=tf.float32)
+
+        x = tf.convert_to_tensor(x, dtype=self.input_dtype)
         x = tf.expand_dims(x, axis=0)
+
+        if self.transform_inp:
```
adelavega, line 171: So you're just saying that this functionality should have been copied over from the parent `TFHubExtractor` class but wasn't?

Is there a way to do this with `super` to make this cleaner/more efficient?

adelavega (changes requested):
I don't think I have any objections in principle just a few minor comments

pliers/extractors/models.py
anibalsolon: I believe it should- it provides some flexibility to adapt the input to the different requirements of TFHub models.

Using super would require the `x` to be a subclass of `Stim` given it uses the accessor `.data`, but it is a TF tensor at this point and it may be not too performatic to do the conversion.

adelavega:
Cool, okay two more things:

1) @rbroc can you take a quick look

2) We need to fix tests before we can merge! I'm guessing its unrelated but lets fix in a separate PR

rbroc:
hey thanks for spotting this and sorry for the delay! 

Re: the dtype problem. Couldn't we simply convert x to tensor before rescaling and casting to `tf.float32` by default, triggering a warning instead of an error?

Re: ```transform_inp```. The current logic leaves this attribute to ```None``` for the image extractor and it remains unused. Not super elegant, but the idea is that in most cases the only meaningful preprocessing transformations for images will be rescaling and resizing.
If we have reasons to think other transformations are relevant, we could allow user to pass ```transform_inp``` to this extractor too. In this case though I think ```transform_inp``` should either be applied to stim.data before resizing or between resizing and rescaling.

anibalsolon:
It may bother a bit the user to keep showing that warning about shapes, but it should be fine. I've *finally* merged your PR, thank you for that!

adelavega:
Hey @anibalsolon 

Do you want to meet about this sometime soon? I'm happy to give this a full review. 

Also, can you rebase w/ master? I've fixed the tests recently

adelavega:
I've looked this over and it looks good to me, and tests are passing so will merge soon.

adelavega (approved):

adelavega:
Test failures are unrelated, so merging

PR #477, opened 2022-09-09
FIx Tests

adelavega:
Due to many upgrades (in particular Librosa), many tests are failing due to slightly different values. 

I'm relaxing many tests to prevent this from happening so frequently.

NLTK also requires more granular resource management now.


adelavega:
Hallelujah!

PR #484, opened 2022-10-27
[enh] support non-English transcript with RevAI

### LINKED ISSUE #479: Allow for non-English transcription with Rev.AI
Currently, it's not obvious how to request languages other than English when using the `RevAISpeechAPIConverter`. From [their docs](https://docs.rev.ai/api/asynchronous/reference/#operation/SubmitTranscriptionJob!ct=application/json&path=language&t=request) it's clear that requests can include a number of languages (including French, which is my use case !), but it's not obvious how to add that param in a pliers call.

From my understanding, [the transcription option will not automatically infer the language](https://docs.rev.ai/resources/tutorials/transcribe-audio-automatic-language-identification/#technical-approach); instead, that should be supplied by the user or done in a first-pass language identification step.

I believe `RevAISpeechAPIConverter` would need to accept a new `language` parameter that could then be passed here:
https://github.com/PsychoinformaticsLab/pliers/blob/9b10b27e70c3fbb7647eb1c70e031f55d824f3f6/pliers/converters/api/revai.py#L73-L77

Does that sound approximately correct / Would that be of interest ?

emdupre:
Closes #479.

- Adds a new `language` parameter to `RevAISpeechAPIConverter`, taking str values documented in the [API docs](https://docs.rev.ai/api/asynchronous/reference/#operation/SubmitTranscriptionJob!ct=application/json&path=language&t=request).
- Sets a minimum version of the `rev_ai` python SDK
- Implements a smoke test that non-supported languages error on conversion.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ env.list
 docs/_build/
 docs/generated/
 .pytest_cache/
+.vscode/
--- a/optional-dependencies.txt
+++ b/optional-dependencies.txt
@@ -24,6 +24,6 @@ tensorflow>=2.4.0
 torch
 transformers
 xlrd
-rev_ai
+rev_ai >= 2.11.0
 tensorflow-hub
 tensorflow_text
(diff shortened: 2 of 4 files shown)


qmac (approved):
LGTM!

adelavega (approved):

adelavega:
Thanks @emdupre

PR #486, opened 2022-11-07
CI/TESTING: Updated ffmpeg, expand to Python 3.10

adelavega:
ffmpeg failing on remote files in CI


adelavega:
It _should_ work with what's in the channel for ubuntu but im not sure

adelavega:
Hallelujah

